PR #2, opened 2022-08-18
Create MIT LICENSE

ssandino:
Created the MIT License


andrashee (approved):
👍 LGTM

andrashee:
We can play around with something like this https://www.npmjs.com/package/license-checker to check the licences of our dependencies. Will have a look at it when migrating the exisiting projects.

PR #36, opened 2022-09-02
Switch from yarn to npm

mkue:
Due to yarn having performance issues with Docker, we switch back to `npm`.

This is still a draft because I still have these two issues:

- `TypeError: this.libOptions.parse is not a function`
  https://youtrack.jetbrains.com/issue/WEB-57089/ESLint823-TypeError-thislibOptionsparse-is-not-a-function
  
- Some typing problems, no idea why they appeared now...
  ```
  TS2344: Type 'Organisation' does not satisfy the constraint '{ [Key: string]: CMSType; }'.
  ```


.github/workflows/admin-pr.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           firebase setup:emulators:ui
 
       - name: Install Dependencies
-        run: yarn install
+        run: npm install
 
       - name: Run Tests
         run: yarn emulators:test
```
andrashee, line 28: Shall we completely switch for all commands to npm?

andrashee (commented):
Awesome 😎

.github/workflows/admin-pr.yml
mkue: Yes, I would switch everywhere. I updated all files now.

mkue:
This is now finally done. The typing issues I encountered were caused by FireCMS and fixed by another version upgrade.

We use now `npm` instead of `yarn` everywhere.

andrashee (approved):
Yeah very nice 👍. Thanks for looking into this tedious issue

PR #41, opened 2022-09-04
Algolia text search for recipients

andrashee:
- installed the algolia extension https://console.firebase.google.com/u/1/project/social-income-prod/extensions/instances/firestore-algolia-search?tab=config which creates a firebase function listening on firestore collection changes and synces those to algolia. https://www.algolia.com/apps/T72XPOVPO9/explorer/browse/recipients?page=1&searchMode=search
- import the already existing recipients using the command line tool
- setup https://www.algolia.com/apps/T72XPOVPO9/explorer/browse/recipients?page=1&searchMode=search to have the following fields in the search: email, firstname, lastname, communication phone, mobile money phone
- setup the code to use the algolia search backend. Only works in production, not in local emulator

Note:
- the extension configuration is a bit weird since it seems to allow to index only 1 collection. Not sure if we can retrieve somehow the plain code of the firebase function so that we can duplicate it for other collections. For now it just works for recipients / paymentconfirmations.


ssandino (approved):
Nice!

PR #67, opened 2022-09-16
Migrate postfinance balance function

andrashee:
@mkue could you have a view at the code? I have very limited knowledge of typescript and its best practices. 

I used now a simple include to include the shared interfaces. Do you think this a good way? Tried references as well, but couldn't figure out how to make all the js files available in one output folder.


backend/functions/tsconfig.json
```diff
@@ -10,5 +10,5 @@
     "target": "es2017"
   },
   "compileOnSave": true,
-  "include": ["src", "tests"]
+  "include": ["src", "tests", "../../shared/interfaces/src"]
```
andrashee, line 13: I implemented it now with a simple include statement. 
Like this, the final output directory will contain all the required js files.

image

Feels a bit hacky since we don't have a dedicated package json for the shared directory defining the dependencies  for the shared code. But for now, it's a start.

mkue: This is totally fine for the moment, imo. We should at some point create a separate package where we define all the types, then we won't have these ugly imports.

NPM now also supports workspaces (https://docs.npmjs.com/cli/v7/using-npm/workspaces), so I think it should pretty straightforward to set this up a local package. I'll have a look at this soon.

mkue (approved):
Very cool! All looks good to me 👍

mkue:
There will be a conflict with my branch because you still use `yarn` here, still. Should we merge mine first and then you update this branch to use npm as well? I think you only need to remove both `yarn.lock` and `package-lock.json` files and do a fresh `docker compose backend npm install`.

andrashee:
Top! Merged yours and will adapt this one

andrashee:
Removed the corresponding function from the website repo.

PR #70, opened 2022-09-16
Fonts folder

ssandino:
fonts and license (special remark also in readme and license)


ssandino:
we could also rename /shared to /common if needed (no hard feeling here)

andrashee (approved):
Both common and shared works for me

PR #93, opened 2022-09-24
Data seed with more dummy data

ssandino:
- added more dummy data


ssandino:
@andrashee added more dummy data, but wasn't exactly sure what changed files to include. lets see if this works.

andrashee (approved):
Looks good to me.

PR #99, opened 2022-09-28
[ui] Add typography foundations

renestalder:
Add very basic typography documentation and a component for applying the stack.

Out of scope but planned: Applying variable font sizing (`clamp`) to the predefined typography stack.

**Note**: These sizes are not taken from the old website. We use the predefined sizing stack of Tailwind here, for the sake of standardization.


andrashee (approved):
🎉

PR #100, opened 2022-09-28
Removing fonts

ssandino:
- no longer needed as they are used on a sub-level of each tool
- the SUB font was not used anyway
- licenses are staying in the folder as they count for all different applications


ssandino:
As the UI is referring to the fonts, let's strange strategy: I update the fonts (as 2/3 are the wrong ones) instead of deleting them.

PR #101, opened 2022-09-28
Font amendmend

ssandino:
- erasing two subsets of the font
- adding the main font files
- changing the path in the UI code


renestalder (approved):
lgtm

PR #103, opened 2022-10-01
Restructure project and setup admin and backend as npm workspaces

mkue:


mkue:
This PR turned out to be much bigger than I initially anticipated🙈

I think the project structuring and the way we set up the npm workspaces makes sense now and is consistent. Let me know if you disagree with any restructuring that I made.

@renestalder: I'm encountering some weird issue with storybook and its webpack dependency that I don't know how to solve best. I installed webpack 5 in the admin project. Storybook seems to have some quite strange dependency to webpack 4 that is broken now. As far as I understand it, storybook should still use webpack 4 (it's also installed, see package-lock.json) but somehow it doesn't work. Using [@storybook/builder-webpack5](https://www.npmjs.com/package/@storybook/builder-webpack5) could/should resolve this problem but I saw that you use vite as the builder in the `ui` project and I don't think its compatible with that. Do you have any idea how we can solve this?

@andrashee, let's look at the deployment together and Tuesday and make sure everything still works.

Once this is ready to be merged, I'll obviously squash it because I had to make a lot of commits to get the GH actions working again.

renestalder:
> @renestalder: I'm encountering some weird issue with storybook and its webpack dependency that I don't know how to solve best. I installed webpack 5 in the admin project. Storybook seems to have some quite strange dependency to webpack 4 that is broken now. As far as I understand it, storybook should still use webpack 4 (it's also installed, see package-lock.json (`node_modules/@storybook/core-common/node_modules/webpack`)) but somehow it doesn't work. Using [@storybook/builder-webpack5](https://www.npmjs.com/package/@storybook/builder-webpack5) could/should resolve this problem but I saw that you use vite as the builder in the `ui` project and I don't think its compatible with that. Do you have any idea how we can solve this?

@mkue 

Storybook is using Webpack to build the UI interface of Storybook itself. Vite is used to build our UI components.

One problem could be that you set the `--workspace` flag while installing the admin interface. You probably just should make an `npm ci` from the root to ensure the dependencies can correctly resolve over all workspaces. Also make sure to delete all packate-lock.json files in the newly added workspaces. In the workspaces setup, we only need one package-lock.json in the root that tracks all the dependencies. Use `npm ci` whenever possible to ensure we all have the same dependencies in our development environments (`npm install` can update the lock file, `npm ci` (clean install) doesn't).

Then let us check again if this resolves the issue.

I often switch to `pnpm` (https://pnpm.io/) in monorepos as it's more reliable and faster for tracking these dependencies.

mkue:
I tried all options and orders to install the dependencies and deleted the `node_modules` folder and the `package-lock.json` file about a hundred times😬

I just now stumbled across [this issue](https://github.com/storybookjs/storybook/issues/18696). It's exactly the problem that we face as well, I believe.

What would you do to resolve this? What's the advantage of vite over webpack? Should we also switch to webpack as the builder?

renestalder:
Maybe we can test this solution here:
https://github.com/storybookjs/storybook/issues/18696#issuecomment-1260433125

I'm mostly using Vite because it's a modern solution. I just know what it does under the hood and know how it deals with newer EcmaScript stuff. It's a HTML first solution, and it's just really fast. If you can switch, I recommend it, just for maintainability. All of us that ever maintained a Webpack config over multiple major versions know, that this is a major pain. You will benefit from the better speed, easy configuration, better predictability over your JavaScript chunks (at least that's what it is for me).

mkue:
Cool, that's what I would have tested as well. I thought we had to completely switch to `webpack` as the builder but then realized that this isn't the case.

Everything works now🎉

@andrashee  & @renestalder, ptal and let me know if you disagree with any of my restructurings.

admin/src/collections/Recipients.tsx
```diff
@@ -1,16 +1,16 @@
 import {
```
andrashee, line 1: nit: .ts instead of tsx?

admin/src/collections/index.ts
```diff
@@ -0,0 +1,17 @@
+import { EntityReference } from '@camberi/firecms';
+
+export interface BuildCollectionProps {
+	isGlobalAdmin: boolean;
+	organisations?: EntityReference[];
+}
+
+export { adminsCollection } from './Admins';
```
andrashee, line 8: Out of curiosity: what is the advantage of exporting the collections again here (and for the interfaces) instead of just importing them directly when using them? To have an easy overview of the available ones?

backend/firebase.json
```diff
@@ -1,9 +1,7 @@
 {
 	"functions": {
 		"ignore": ["node_modules", ".git", "firebase-debug.log", "firebase-debug.*.log"],
-		"predeploy": ["npm --prefix \"$RESOURCE_DIR\" run build"],
-		"source": "functions",
-		"codebase": "backend"
+		"source": "."
```
andrashee, line 4: 👍

docker-compose.yaml
```diff
@@ -13,8 +13,8 @@ services:
       - "8080:8080" # firestore emulator
       - "9199:9199" # storage emulator
       - "4000:4000" # emulator UI
-    working_dir: /home/node/admin
-    command: /bin/bash -c "npm install && npm run emulators:start"
+    working_dir: /home/node/
```
andrashee, line 16: Given the makefile: shall we still have the different docker-compose services?
Or should we just have one, with all ports exposed, and then override the run command in the make files?

andrashee (approved):
Amazing work @mkue! Thanks so much for going through this. Feels like a massive step to me to get towards a homogenous, modular and maintainable code base.

LGTM

andrashee:
> **Unrelated**: Shouldn't we store the package related types, e.g. the admin types, in the admin project instead of the shared folder (as they are admin specific, even though they might be used somewhere else)? As we now use workspaces in all projects, I would assume I just cross-reference the types from another workspace package by installing the package as dependency in the package that consumes them (via npm instead of relative paths).

Is your suggestion to have the interfaces in admin and add the admin module as dependency to the backend module? Would that drag in all the dependencies of the admin (e.g. firecms) to the backend? I think that would be unfortunate.

andrashee:
> Visit the preview URL for this PR (updated for commit [df8a997](https://github.com/socialincome-san/public/commit/df8a99765c51093cf5e6fb5867b37413f50fc655)):
> 
> https://admin-social-income-prod--pr103-mkue-admin-setup-sha-oaz70lck.web.app
> 
> (expires Mon, 10 Oct 2022 11:01:56 GMT)
> 
> 🔥 via [Firebase Hosting GitHub Action](https://github.com/marketplace/actions/deploy-to-firebase-hosting) 🌎

This doesn't work anymore. It seems to be in emulator mode instead of production mode. Will have a look at it in a separate PR.

docker-compose.yaml
renestalder: I guess it makes sense to still keep the different Docker compose services. I only see Makefiles as a way to make it easier to work with Docker, but on an application architecture level, it makes sense to me that they are different services in the compose file considering they are also different services on the end-server.

renestalder:
> > **Unrelated**: Shouldn't we store the package related types, e.g. the admin types, in the admin project instead of the shared folder (as they are admin specific, even though they might be used somewhere else)? As we now use workspaces in all projects, I would assume I just cross-reference the types from another workspace package by installing the package as dependency in the package that consumes them (via npm instead of relative paths).
> 
> Is your suggestion to have the interfaces in admin and add the admin module as dependency to the backend module? Would that drag in all the dependencies of the admin (e.g. firecms) to the backend? I think that would be unfortunate.

You are right, that's probably not what we want. I didn't think of that. It's just clear to me that I will reference the UI as a dependency in the website project later and that all UI types live in the UI package, so that's why it seemed strange to me that the admin types are so far away from its main package.

renestalder (approved):

.github/workflows/admin-pr.yml
```diff
@@ -23,13 +20,13 @@ jobs:
...
-        run: npm install
+        run: npm --workspace=@socialincome/admin install
 
       - name: Run Tests
-        run: npm run emulators:test
+        run: npm run admin:test
 
       - name: Build
-        run: npm run build
+        run: npm run admin:build
```
andrashee, line 29: @mkue: I think overriding the env vars doesn't work any more. That's the reason the preview (and then the final deployment, I assume) is broken.

Might be related to the dotenv config?
```
		new Dotenv({
 			path: path.resolve(path.join(__dirname, '.env')),
 		}),
```

We can also change the way how we set the production variable. The current setup is not nice:
a) we have code duplication in admin-pr and admin-deployment github workflow
b) we nullify the auth envs here

In the backend deployment, I write the github secrets to a .env file. We could do the same here, then your dotenv config should work again.

We could remove the `admin-deployment` workflow and use the `admin-pr` and in the deployment step have an if branch==main => channelId: live.

Wdyt?

.gitignore
```diff
@@ -27,6 +27,3 @@ build
 # JetBrains
 .idea/
 .ignore/
-
-admin/social-income-prod-firebase-adminsdk.json
```
andrashee, line 31: Shall we keep this, so that no one accidentally commits this? I think it's currently used to run the scripts, right?

.github/workflows/admin-pr.yml
andrashee: @mkue, I added a few commits for the github workflows. Hope that's fine. The preview works now again, and we only have one yaml for PR and deployment.

mkue: super nice, thanks!

.gitignore
mkue: Yes, let's keep it👍🏼

In the future, the scripts should also run against the local firestore imo. We could integrate some buttons into the admin UI to trigger them in production, what do you think?

admin/src/collections/index.ts
mkue: Yes, I think that was my intention :) We can also use direct imports, I have no strong preference...

mkue:
I think we should adapt and extend the types so that they are not only compatible with FireCMS, i.e. allowing not only `EntityReference` as a type for referencing objects but also `DocumentReference` which is the type used in the "firestore" package. Then, we should be able to reuse the types in all projects.

admin/src/collections/Recipients.tsx
mkue: It has to be `.tsx`. The `AsyncPreviewComponent` uses JSX syntax.

mkue:
Thanks for adapting the GH actions @andrashee. I'll merge this now. If anything breaks, we can fix it tonight :)

andrashee:
👍 

image

https://github.com/socialincome-san/public/actions/runs/3180589804/jobs/5184347268

I think you might need to add the main again in the backend package.json: `"main": "lib/backend/functions/src/index.js",`

mkue:
> 👍
> 
> image
> 
> https://github.com/socialincome-san/public/actions/runs/3180589804/jobs/5184347268
> 
> I think you might need to add the main again in the backend package.json: `"main": "lib/backend/functions/src/index.js",`

Let's quickly have a look at this together later today👍🏼

Issue #104, opened 2022-10-02
[Admin] Adding referral field to user collection


ssandino:
**Is your feature request related to a problem? Please describe.**
We currently collect the information from new contributors, how they learned about Social Income, but it's not yet visible in the admin tool

**Describe the solution/feature** 
Adding the field into the admin view

ssandino:
see #105

PR #105, opened 2022-10-02
Adding referral field #104

ssandino:
- added it in the user collection


mkue:
Lgtm👍🏼 but given the fact that I refactored how we use the types from the shared directory, I think it makes sense to integrate these changes into my PR (#103). I'll add it next week.

mkue:
Done in #135

PR #109, opened 2022-10-04
fix(ui): only trigger workflow for UI related changes

andrashee:
I think the workflow has at the moment the following issues:
- it runs for unrelated PRs. E.g. https://github.com/socialincome-san/public/pull/106
. It also runs the action twice in PRs (once for the pull request trigger, once for the push) E.g. https://github.com/socialincome-san/public/pull/103

With this change, only pushes to main or PRs are triggers. For both, the path filter is used.


.github/workflows/ui.yml
```diff
@@ -1,17 +1,18 @@
 name: ui - build and deploy
 
 on:
+  pull_request:
+    paths:
+      - "ui/**"
+      - "package-lock.json"
+      - ".github/workflows/ui.yml"
   push:
+    branches:
```
renestalder, line 10: @andrashee 

> only pushes to main or PRs are triggers

I'm pretty sure this is not what it does. Based on the workflow documentation of GitHub, the `push` rule here will now run **when at minimum one of both definitions in the push rule applies** (not _when at least both apply_):

- When someone pushes to main
- When there are changes in the defined paths

Based on that, this rule change would make that any changes on the main branch (also in the admin folder or on the readme) always trigger the workflow.

Prove me wrong, but that was actually what I always found when I looked up how to make this an **AND** rule.

andrashee: The
```
branches:
   - main
paths:
   - xyz
```
is definitely an AND condition. We are using the same pattern in the other workflows.

The documentation is not so clear what is chained as an OR or AND. Here someone had the same question: https://stackoverflow.com/questions/63460875/github-actions-how-do-you-trigger-a-push-when-a-specific-directory-in-a-branch

Top-level (pull_request, push) are ORs.

andrashee: @renestalder, wdyt shall we try this?

renestalder: :+1:

renestalder (approved):

PR #115, opened 2022-10-04
feat(shared): update README.md and split into multiple ones

andrashee:
- Moved the individual instructions into the submodules to keep the main readme more concise.
- Adapted commands to makefile
- Still needs proofreading.


Makefile
```diff
@@ -5,11 +5,14 @@ admin-build:
...
 
 admin-serve:
-	docker compose up admin
+	docker compose run --rm admin bash -c "npm --workspace=@socialincome/admin install && npm run admin:serve"
 
 admin-test:
 	docker compose run --rm admin bash -c "npm --workspace @socialincome/admin install && npm run admin:test"
 
+admin-export-seed:
+    docker exec -it public-admin-1 npm run emulators:export
```
renestalder, line 14: Are you sure that works in every case with the static public-admin-1 name here? I think it could be possible that the container has different naming/number count on other computers depending on how docker compose created it in the first place (e.g. naming conflict). Could we achieve the same with a docker compose run command?

ssandino (approved):
splitting up main readme file is fine with me

ssandino:
There seems to be a traffic jam in updating the readme file. But all straightforward conflicts.

Makefile
```diff
@@ -5,11 +5,14 @@ admin-build:
 	docker compose run --rm admin bash -c "npm --workspace @socialincome/admin install && npm run admin:build"
 
 admin-serve:
-	docker compose up admin
+	docker compose run --rm admin bash -c "npm --workspace=@socialincome/admin install && npm run admin:serve"
```
mkue, line 8: If you use `run` you need to add the option `--service-ports`, otherwise, the ports to the container are not open.

Makefile
mkue, line 14: Why do you use exec instead of run here?

mkue (approved):
lgtm

ssandino:
@andrashee should we go on with this? Lgtm. The outstanding questions we could also address outside of the issue. With a concise main readme file it will be easier to communicate with the open source community.

mkue:
Before we merge this, `backend` should be renamed to `functions` now.

MDikkii (approved):

andrashee:
@ssandino updated it to reflect the latest refactorings

Makefile
andrashee: Good point! Changed it to docker compose exec.

PR #119, opened 2022-10-07
wip: feat(shared): make checks required

andrashee:
Currently, we only trigger the checks when a file changes in the relevant module to not run them unnecessarily and slow down the merge process.

However, there is one problem with that. We can't make the checks required since they are not always triggered. This is also explained in https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks. 
The github suggests a workaround by having another dummy workflow which always triggers, but that feels unnecessarily verbose.

I'm trying out this https://github.com/marketplace/actions/skip-duplicate-actions plugin which should address it as well.


.github/workflows/admin.yml
```diff
@@ -2,14 +2,14 @@ name: Admin - build and deploy
 on:
   pull_request:
     paths:
-      - "admin/**"
+      - "backend/**"
```
mkue, line 5:
```suggestion
      - "functions/**"
```

.github/workflows/admin.yml
```diff
@@ -2,14 +2,14 @@ name: Admin - build and deploy
...
-      - "admin/**"
+      - "functions/**"
       - "shared/src/**"
       - ".github/workflows/admin.yml"
   push:
     branches:
       - main
     paths:
-      - "admin/**"
+      - "backend/**"
```
mkue, line 12:
```suggestion
      - "functions/**"
```

mkue:
I'm confused, what happened to the plugin that you tried out?

andrashee:
> I'm confused, what happened to the plugin that you tried out?

Didn't trigger the workflow at all. Will have another look at this later today.

PR #123, opened 2022-10-07
Make UI package available for usage in other projects

renestalder:
- Adds Rollup build to create esm, commonjs and typescript typings
- Adds script to automatically create import/export barrels
- Update README with documentation about how to setup and use components in a project


andrashee (approved):
Nice 👍

PR #124, opened 2022-10-08
refactor (functions): restructure project setup

mkue:
@andrashee, I think this resolves most of the issues we discussed. Let me know what you think of the new structure.

I think a single `firebase.json` file makes sense because everything we deploy is in a single Firebase project. We should switch to multiple `firebase.json` files once we have more than one active project.

FYI, I switched back to using `tsc` for building the functions because I got the following error when I tried to deploy the functions that were build with webpack. 
```
'@socialincome/shared@^0.1.0' is not in this registry
```
I removed the `@socialincome/shared` dependency from the `package.json` and import the code with relative paths again. Like this, it works at least.

The `firebase deploy` command does not seem to like monorepos with npm workspaces. We should keep an eye on this issue: https://github.com/firebase/firebase-tools/issues/653.


mkue:
I would leave the webpack config in the `functions/` directory in this PR. We can remove at a later point. Like this, we at least have an example of a working config.

mkue:
Also, we get this annoying warning in the admin front end now:

Screen Shot 2022-10-08 at 16 16 51

The [issue](https://github.com/firebase/firebase-js-sdk/issues/6660) is already resolved and the error should go away with the next firebase release.

admin/src/CallDummyFunctionButton.tsx
```diff
@@ -0,0 +1,26 @@
+import { PlayArrow } from '@mui/icons-material';
+import { IconButton, Tooltip } from '@mui/material';
+import { getFunctions, httpsCallable } from 'firebase/functions';
+
+export default function CallDummyFunctionButton() {
```
andrashee, line 5: 🥳

admin/src/config.tsx
```diff
@@ -1,21 +1,23 @@
...
 export const FB_STORAGE_EMULATOR_PORT = process.env.REACT_APP_ADMIN_FB_STORAGE_EMULATOR_PORT;
+export const FB_FUNCTIONS_EMULATOR_HOST = process.env.REACT_APP_FB_FUNCTIONS_EMULATOR_HOST;
+export const FB_FUNCTIONS_EMULATOR_PORT = process.env.REACT_APP_FB_FUNCTIONS_EMULATOR_PORT;
 
 // algolia search
 export const ALGOLIA_APPLICATION_ID = process.env.REACT_APP_ADMIN_ALGOLIA_APPLICATION_ID;
 export const ALGOLIA_SEARCH_KEY = process.env.REACT_APP_ADMIN_ALGOLIA_SEARCH_KEY;
+
+// production envs
+export const FIREBASE_CONFIG = {
```
andrashee, line 15: 👍

functions/package.json
```diff
@@ -0,0 +1,40 @@
...
+	"name": "@socialincome/functions",
+	"private": true,
+	"engines": {
+		"node": "16"
+	},
+	"main": "dist/functions/src/index.js",
+	"scripts": {
+		"build": "tsc",
+		"build:watch": "tsc --watch",
+		"build:webpack": "webpack --config webpack.config.js",
```
andrashee, line 11: for what are we requiring now the webpack? For the final deployment?

functions/src/etl/importPostfinanceBalance.ts
```diff
@@ -20,6 +20,7 @@ export const importBalanceMailFunc = functions.pubsub.schedule('0 * * * *').onRu
 
 export const retrieveBalanceMails = async (): Promise<BankBalance[]> => {
 	try {
+		functions.logger.info('Updated function');
```
andrashee, line 23: remove again?

shared/package.json
```diff
@@ -3,6 +3,6 @@
 	"version": "0.1.0",
 	"private": true,
 	"devDependencies": {
-		"@camberi/firecms": "2.0.0-alpha.30"
+		"@camberi/firecms": "2.0.0-alpha.37"
```
andrashee, line 6: I think it would be nice if the shared package doesn't have a dependency to the firecms.

What do you think of the approach I had in https://github.com/socialincome-san/public/pull/86/files#diff-eb15213d223e369551834a6ab5f0831a9a4484dfe9a44dabb2e29f91898f84ac where I copied the EntityReference implementation into the shared package? It's the only dependency we use, and then we can get rid of this one.

Makes the shared and therefore functions package much slimmer and faster.

andrashee (commented):
Awesome! Thanks a lot for the patience on this one.

One thing which I don't fully understand yet. Why do we need the webpack packages when we do it with local references and not using the package?

andrashee:
> I would leave the webpack config in the `functions/` directory in this PR. We can remove at a later point. Like this, we at least have an example of a working config.

Ah, haven't seen this comment during my review. Make sense 👍

shared/package.json
mkue: But isn't that a problem if the dependent object changes..? Should we define it as a `peerDependency` instead?

mkue: Hmm, but isn't that a problem when the dependent object changes? If it's just a `devDependency`, I don't see the problem. It will be in your `node_modules` folder anyway because it's being used in another project.

andrashee: Fair enough. Feels conceptually wrong to me that the functions have a compilation dependency to the firecms, but we can still iterate over it later.

```
npm ERR! Error: Missing script: "functions:test"

To see a list of scripts, run:
  npm run 
npm ERR!   in workspace: @socialincome/admin@0.1.0 
npm ERR!   at location: /home/node/admin 
npm ERR! Missing script: functions:test

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/node/.user/.npm/_logs/2022-10-08T15_06_34_214Z-debug-0.log
make: *** [functions-test] Error 1

~/code/public restructure-functions *1 ?2 ❯ make functions-test   
```
should that work?

andrashee: Sorry my bad cached container.

andrashee (approved):
very nice!

PR #127, opened 2022-10-09
feat(shared): make checks required - second attempt

andrashee:
Currently, we only trigger the checks when a file changes in the relevant module to not run them unnecessarily and slow down the merge process.

However, there is one problem with that. We can't make the checks required since they are not always triggered. This is also explained in https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks.
The github suggests a workaround by having another dummy workflow which always triggers, but that feels unnecessarily verbose.

A million commits later, I'm proposing to use this https://github.com/marketplace/actions/skip-duplicate-actions plugin which should address it as well.

I sneaked in another change. @mkue running the tests for the admin we go the following errors when starting the firebase emulators:
image

This is because we need the functions now to be built, since they are a dependency of the admin. Building it now before running the tests. Wdyt @mkue?
Then the output looks like
image

Next step is then to mark all checks as required in the Github Settings

### DIFF
--- a/.github/workflows/admin.yml
+++ b/.github/workflows/admin.yml
@@ -32,10 +32,7 @@ jobs:
           firebase setup:emulators:ui
 
       - name: Install Dependencies
-        run: |
-          npm --workspace @socialincome/functions install \
-          npm --workspace @socialincome/shared install \
-          npm --workspace=@socialincome/admin install
+        run: npm --workspace=@socialincome/admin install
 
       - name: Run Tests
         run: npm run admin:test
--- a/.github/workflows/admin.yml
+++ b/.github/workflows/admin.yml
@@ -1,8 +1,6 @@
 name: Admin - build and deploy
 on:
   pull_request:
-    branches:
-      - main
   push:
     branches:
       - main
@@ -16,14 +14,11 @@ jobs:
       - id: skip_check
         uses: fkirc/skip-duplicate-actions@v5
         with:
-          paths:
-            - "admin/**"
-            - "shared/src/**"
-            - ".github/workflows/admin.yml"
+          paths: '["admin/**", "shared/src/**", ".github/workflows/admin.yml"]'
 
   test_deploy:
     needs: check_applicability
-    if: needs.pre_job.outputs.should_skip != 'true'
+    if: needs.check_applicability.outputs.should_skip != 'true'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo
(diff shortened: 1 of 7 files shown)


.github/workflows/admin.yml
```diff
@@ -37,7 +32,10 @@ jobs:
           firebase setup:emulators:ui
 
       - name: Install Dependencies
-        run: npm --workspace=@socialincome/admin install
+        run: |
+          npm --workspace @socialincome/functions install \
```
renestalder, line 36: Nitpik: At this point I wonder if it wouldn't be better to just run 'npm ci' (installs all workspaces) at all times and cache the .npm folder correctly. I think that is probably faster than 3 separate installs.

renestalder (approved):

.github/workflows/admin.yml
mkue: @andrashee, why did you have to change this at all? Shouldn't installing the admin dependencies be enough?

Makefile
```diff
@@ -8,7 +8,7 @@ admin-serve:
 	docker compose up admin
 
 admin-test:
-	docker compose run --rm admin bash -c "npm --workspace @socialincome/admin install && npm run admin:test"
```
mkue, line 11: Same here...

package.json
```diff
@@ -11,8 +11,8 @@
 	"scripts": {
 		"format-code": "prettier --write .",
 		"admin:build": "$npm_execpath --workspace=@socialincome/admin run build",
-		"admin:serve": "$npm_execpath --workspace=@socialincome/admin run serve",
-		"admin:test": "$npm_execpath --workspace=@socialincome/admin run test",
+		"admin:serve": "$npm_execpath run functions:build && $npm_execpath --workspace=@socialincome/admin run serve",
```
mkue, line 14: 👍

mkue (approved):

mkue:
I think it crashes because there is no `functions/dist/functions/src/index.js` file that the emulator tries to load. At the moment, we don't need the functions to be up and running for the admin tests.

Couldn't we just change the test script in `admin/package.json` to this (add `--only firestore`)?
```
firebase emulators:exec --project demo-social-income-prod --only firestore --config ../firebase.json 'npm run test:app'
```

If that doesn't work, the way you did it also looks good to me👍

andrashee:
> I think it crashes because there is no `functions/dist/functions/src/index.js` file that the emulator tries to load. At the moment, we don't need the functions to be up and running for the admin tests.
> 
> Couldn't we just change the test script in `admin/package.json` to this (add `--only firestore`)?
> 
> ```
> firebase emulators:exec --project demo-social-income-prod --only firestore --config ../firebase.json 'npm run test:app'
> ```
> 
> If that doesn't work, the way you did it also looks good to me👍

You are right, for the running the admin tests, we don't need the running functions. Good suggestion, let me limit it to start the firestore only.

.github/workflows/admin.yml
andrashee: Not required any more when we remove the functions dependencies from the tests 👍

Issue #55, opened 2022-09-11
Update Readme with profiles of code contributors


ssandino:
**Goal:** 
Display the GitHub contributors right in our Readme to create a strong(er) sense of community. 

Good example: [Read me of Contributor Covenant](https://github.com/EthicalSource/contributor_covenant/blob/release/README.md) and their embedded [SVG](https://opencollective.com/contributor-covenant/contributors.svg?width=890&button=false)

**Why is it a good example?** 

- It scales nicely
- profile images are sharp, round and small
- it can be embedded and therefore done with the straightforward markdown language

There are many tools doing it (like [this](https://github.com/dummy-crew/svg-contributors)). Open questions: which one to implement, how to automatically save the SVG on our own server (as it could also be used on our website) and how do new contributors over GitHub appear automatically?

So visually it should be exactly like [this](https://github.com/EthicalSource/contributor_covenant/blob/release/README.md) (small circles).

daksh10110:
Hey I can do this, can you assign this task to me.

Yatharth-Jain:
Hey There, I would like to Contribute in this Task. Along with that this is my first Open Source step so also want some guidance. So can you assign that task to me and guide me on any issue

ssandino:
> Hey I can do this, can you assign this task to me.

how is it going @daksh10110 – anything we can help?

daksh10110:
I am extremely sorry for not informing before, I had a lot of work lately so I wasn't able to do this task.

I have my exams from the next week so it would take some time, you can assign it to some other person if you wish to.

The idea I had in mind was to get the photos using GitHub api and then using flask and pillow to send the files from there.

ssandino:
Thanks for the quick answer and sharing your approach. Hope your exams went well. I also assign @Yatharth-Jain in case he is still interested and wants to push this issue forward. Then we see how to wrap it up.

ssandino:
If we make this feature more versatile, we open up a new issue

PR #131, opened 2022-10-12
[Shared] Update Readme with profiles of code contributors #55

Yatharth-Jain:
#55

Added Contributors Pictures in the contribution portion of the readme as shown in the picture
image

Kindly give a visit too for reconfirming 
Also can you provide me with the content to type in the Same portion of the readme


ssandino:
@Yatharth-Jain thanks. We were aware of [contrib.rocks](https://github.com/lacolaco/contributors-img/), however, it didn't match our criteria. We would like to have something that is: 

- dynamic (it updates regularly by itself)
- and uses the SVG instead PNG format (sharper)

I think [this tool](https://github.com/dummy-crew/svg-contributors) could do the job. The last question would then be where to save it (whenever there is a change). A quick solution could be to save the generated SVG in a publicy accessible folder on a Google Drive (overwrite last file, so the referenced file in readme is always the newest).

ssandino (changes requested):
@Yatharth-Jain thanks. We were aware of [contrib.rocks](https://github.com/lacolaco/contributors-img/), however, it didn't match our criteria. We would like to have something that is:

dynamic (it updates regularly by itself)

- and uses the SVG instead PNG format (sharper)
- I think [this tool](https://github.com/dummy-crew/svg-contributors) could do the job. 

The last question would then be where to save it (whenever there is a change). A quick solution could be to save the generated SVG in a publicy accessible folder on a Google Drive (overwrite last file, so the referenced file in readme is always the newest).

Yatharth-Jain:
@ssandino thankyou for responding
i would like to ask few queries

Yatharth-Jain:
@ssandino thank you for responding
I will try my best to achieve that .
there somethings i want to ask

first of all i guess the images from the [contrib.rocks](https://github.com/lacolaco/contributors-img/) which we fetch with the help of Img tag in readme should be updated if it create the image on request.

secondly if i have to make new image on every changes to the list of contributors then for that i have to make a scraper which will check for any change, thats my solution would you like to suggest some

thirdly the Repo you have given ,gave glitched svg when i used it as shown below

contributors

any solution for that

ssandino:
Hi @Yatharth-Jain, thanks for the updated PR and good idea with updating it with a regular GitHub action (as you do know via another public repo). Looks good to me. 

@mkue @andrashee good for you? I would propose to deploy this PR and then see further if we can integrate the external GitHub action (see [here](https://github.com/Yatharth-Jain/Social-Income-contributors-list/blob/master/.github/workflows/actions.yml)) into our own workflow (to have not a dependency to another repo, but no risk so far as it's just importing one image that could fail)?

andrashee:
Awesome! Will have a look at it tomorrow morning 😊

andrashee:
> @Yatharth-Jain thanks. We were aware of [contrib.rocks](https://github.com/lacolaco/contributors-img/), however, it didn't match our criteria. We would like to have something that is:
> 
> * dynamic (it updates regularly by itself)

This is the case with @Yatharth-Jain's first suggestion. When you use `https://contrib.rocks/image?repo=socialincome-san/public` as img src it should get updated.

> * and uses the SVG instead PNG format (sharper)

True, it's not super crispy.

andrashee:
> @mkue @andrashee good for you? I would propose to deploy this PR and then see further if we can integrate the external GitHub action (see [here](https://github.com/Yatharth-Jain/Social-Income-contributors-list/blob/master/.github/workflows/actions.yml)) into our own workflow (to have not a dependency to another repo, but no risk so far as it's just importing one image that could fail)?

Is the generated SVG better than the one generated by contrib.rocks?
Here a screenshot how it looks for me
Screenshot 2022-10-14 at 08 28 07

I personally would keep things simple here and use @Yatharth-Jain's first suggestion and just link to contrib.rocks. Having a Github action for this almost sounds a bit like an overkill to me.
E.g. we could add it as markdown to keep things consistent:
```
[![Contributors](https://contrib.rocks/image?repo=socialincome-san/public)](https://github.com/socialincome-san/public/graphs/contributors)

```

If we want to have generate it by ourselves, I would integrate the action into this repository instead of linking to an external one.

Yatharth-Jain:
@ssandino @andrashee My PR is not complete yet as some errors are occuring in the GitHub Actions

And i am currently working on it and may require some time
In the mean time if you can check for the errors which are coming while environment setup of Github action in my Repo, That would be appreciated

ssandino:
I wasn't aware of that contrib.rocks updates automatically. In my opinion, the large avatars generated by contrib.rocks are bit too obtrusive. However, it seems that there is a parameter `columns={number}` [now supported](https://github.com/lacolaco/contributors-img/issues/1110) which may lets us control the size of the avatars (by overwriting the default column count (12) with a higher number). So if this works, let's go with this solution for now – as it takes only a few minutes to get it up and running. 

@Yatharth-Jain feel free to make a second PR which only adds the code (with the addition to make the avatars smaller), so the commit is attributed to you as part of the Hacktoberfest : 
```
[![Contributors](https://contrib.rocks/image?repo=socialincome-san/public)](https://github.com/socialincome-san/public/graphs/contributors)

```

**Why I would then continue to aim and work on an SVG solution:** 
- SVG keeps the information in code which we can better access
- Lets us reuse parts on our website in ways we couldn't with a PNG
- i.e. even introduce optional line breaks in the single SVG holding all avatars 
- i.e. reuse each single avatars on our website by extracting the element (like each avatar floating over the page)

These example applications are out of scope of this issue, but the basic SVG would allow for tackle this in the future.

Yatharth-Jain:
Ok @ssandino I will close this PR and will send a new one right away
Along with that i will look into this problem of svg just for my betterment and once that's finish i will let you know then that your decision to include it or not

ssandino:
@Yatharth-Jain it's deployed. Thanks for looking into the SVG solution.

PR #134, opened 2022-10-12
chore(functions): clean up webpack from functions

mkue:
As we don't need it anymore, let's also remove the config.


andrashee (approved):
👍

PR #135, opened 2022-10-12
chore(admin): Add referral field to User model

mkue:
@ssandino, what do you want to store in the referral field? Is it just a string or is it a reference to something else?

### DIFF
--- a/admin/src/collections/Users.ts
+++ b/admin/src/collections/Users.ts
@@ -62,6 +62,16 @@ const CityCol: AdditionalFieldDelegate<User> = {
 	dependencies: ['address'],
 };
 
+const ReferralCol: AdditionalFieldDelegate<User> = {
+	id: 'referral_col',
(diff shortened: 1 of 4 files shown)


ssandino:
Just the string

ssandino (approved):

mkue:
FYI @andrashee, #134 needs to be merged first

admin/src/collections/Users.ts
```diff
@@ -62,6 +62,16 @@ const CityCol: AdditionalFieldDelegate<User> = {
 	dependencies: ['address'],
 };
 
+const ReferralCol: AdditionalFieldDelegate<User> = {
+	id: 'referral_col',
+	name: 'Referral',
+	builder: ({ entity }) => {
+		let values = entity.values;
```
andrashee, line 69: nit: const?

andrashee (approved):

PR #141, opened 2022-10-14
[shared] Minor setup changes & marking checks required

andrashee:
- Starting from a plain repository, I still got some errors because the dependencies haven't been installed for the shared packages. Switched now to install everything. For me, `npm install` is quicker than `npm ci` since the later deletes the node_moduels every time.

Screenshot 2022-10-14 at 07 40 03

- to make the checks required, we need to set unique names for the jobs, otherwise I can't select them in the UI

### DIFF
--- a/.github/workflows/admin.yml
+++ b/.github/workflows/admin.yml
@@ -1,4 +1,4 @@
-name: Admin - build and deploy
+name: Admin
 on:
   pull_request:
   push:
@@ -7,6 +7,7 @@ on:
 
 jobs:
   check_applicability:
+    name: Check Applicability
     runs-on: ubuntu-latest
     outputs:
       should_skip: ${{ steps.skip_check.outputs.should_skip }}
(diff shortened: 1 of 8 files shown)


renestalder:
It's okay to use `npm install` locally (at least since version 7 or 8 I guess). It's just important that it is `npm clean-install` (`npm ci`) in the GitHub actions.

> For me, npm install is quicker than npm ci since the later deletes the node_moduels every time.

That's right. Just not as it can unexpected alter the package-lock file. But that's okay in our case locally.

andrashee:
> It's okay to use `npm install` locally (at least since version 7 or 8 I guess). It's just important that it is `npm clean-install` (`npm ci`) in the GitHub actions.
> 

Let me change that in the actions.

> > For me, npm install is quicker than npm ci since the later deletes the node_moduels every time.
> 
> That's right. Just not as it can unexpected alter the package-lock file. But that's okay in our case locally.

renestalder (approved):
lgtm. That's great that we now have the full install everywhere because stuff is anyway so much interconnected now that it is easier and saver to just install everything.

andrashee:
Configured that the checks are required

image

PR #143, opened 2022-10-19
NextJs i18n demo

andrashee:
Quick example project to explore a way to implement a multi-language website.

## Stack
There are several components involved:
- [Nextjs & its locale routing](https://nextjs.org/docs/advanced-features/i18n-routing): takes care of handling the routing of the paths. Automatically redirects to the best locale according to the language set in the user's browser. In comparison to the custom implementation in the current website, this doesn't take into account the user's location based on the IP (e.g to show the best currency). Assessment: works fine out of the box, but is relatively basic. One can extend it using middlewares similarly to the current express.js middlewares 
- Translation framework: [react-i18next](https://react.i18next.com) which is a wrapper around [i18next](https://www.i18next.com). Assessment: works very well and comes with many features like pluralization, currency formatting. Seamless between server-side or client rendering. Uses the standard `/locale/{language}/{component}.json` format for the translation files.
- Key extraction for the translation keys. With [i18next-parser](https://github.com/i18next/i18next-parser) we can execute `npm run extract-translations` to automatically extract (and remove unused) translation keys into the translation jsons. Assessment: very handy to keep the translation files in sync with the code. Can be part of the build jobs.
- Online translation tool: [https://www.locize.app/cat/12ao3d8p/v/latest#](locize) can be used to translate the generic json translation files. With `npm run sync-translations` the translations jsons are synced to locize. Missing keys are added in locize and new translations are downloaded and added to the jsons. Assesment: While the UI feels a bit dated, the process of translating new keys and languages feels fast. Locize also has a CDN to directly serve the translation to the website. Like this, changes to translations are always immediately published. However, I would just use the sync command to regularly (e.g. through a github action) update the translation files. Then all the content is in Github and we are only loosely tied to Locize and can switch the tool without any issues.  

In this setup, all components can be individually exchanged. 
E.g. `nextjs` with a different backed, `[react-i18next]` could be replaced with `react-intl` or `Locize` could be replaced with `Lokalise`.

## Workflow

I would imagine the following workflow:
1. a developer adds a new react component with new translation keys.
2. she/he runs `npm run extract-translations` to update the locale jsons.
3. she/he adds the new translation to the English one (or optionally of course also for the other languages).
4. she/he creates a PR.
5. in the PR a github action syncs the translation json with locaize.
6. a (potentially different) translator translates the new translations in locize.
7. after merging, a github action downloads the latest translation jsons from locize and commits them to github
8. the project is built, packed and deployed including the static translation files

## Demo

How to run the demo:
```
cd website/demo
npm install
npm run dev
visit http://localhost:300
```

### Screenshots
Screenshot 2022-10-19 at 22 43 39
Screenshot 2022-10-19 at 22 43 34
Screenshot 2022-10-19 at 22 43 27


renestalder:
Nice.

Couple questions:

* Does this also work without NextJS? As, we should be able to maintain and use translations outside of NextJS, like in the UI components in Storybook. And I'm also fairly certain, we might reuse the translations over multiple other projects.
* That's more NextJS related: As I'm still fairly critical of using an "application" framework for building websites, as for the usual drawbacks in size of what is unnecessarily transferred over the network. How big is the payload at this point when you [statically generate](https://nextjs.org/learn/basics/data-fetching/with-data) this? This should probably be way under 10kb, but I wouldn't be surprised if NextJS adds a 50kb JavaScript bundle and maybe also a JSON payload to hydrate the translations.

I just saw something similar in another project I started to work in. They use `react-intl` and [Accent](https://nextjs.org/learn/basics/data-fetching/with-data) to achieve something similar. To make it work in Storybook I just added the react provider wrapper of react-intl as a global story decorator in the Storybook config. It's mostly the same as your solution (except for the external translation editor), so I find that solution you are proposing great.

ssandino:
Nice setup! Works well locally. 

**2 Thoughts on Search Engine & Language**
(1) I like that Next.js automatically adds the lang attribute to the `<html>`. This also helps to be better displayed on search engines. Now it's a bit messy in the search results as different languages are mixed up. We would then only have to add the additional `hreflang meta tags`.

Currently (Issue: it mixes languages, I thought Google is smarter than this): 
Screen Shot 2022-10-20 at 17 41 52

Nothing that couldn't achieved with some effort on the current website, but nice to have it out of the box. 

(2) With this setup we could also generate a [sitemap at build time ](https://vercel.com/guides/how-do-i-generate-a-sitemap-for-my-nextjs-app-on-vercel)(even runtime). This also helps for SEO and is an issue that has never been solved in the last 3 years. @andrashee does this seem technically doable? 

**1 Thought on Translations**
The workflow makes sense to me. As we have more languages, we'll always end up in the situation, where we have a new page or element ready to deploy, but not yet in all languages. And to be not hold back by translations (which can take time, when we are depending on outside people) it would be nice if we can also deploy with a missing language. In this case, it should just fall back or display the default (EN) until the missing translation keys are added. Is this possible @andrashee ?

andrashee:
> Nice.
> 
> Couple questions:
> 
> * Does this also work without NextJS? As, we should be able to maintain and use translations outside of NextJS, like in the UI components in Storybook. And I'm also fairly certain, we might reuse the translations over multiple other projects.

Absolutely! This is independently of Nextjs. One can do put them into the UI components in Storybook. 

> * That's more NextJS related: As I'm still fairly critical of using an "application" framework for building websites, as for the usual drawbacks in size of what is unnecessarily transferred over the network. How big is the payload at this point when you [statically generate](https://nextjs.org/learn/basics/data-fetching/with-data) this? This should probably be way under 10kb, but I wouldn't be surprised if NextJS adds a 50kb JavaScript bundle and maybe also a JSON payload to hydrate the translations.

It definitely adds an overhead in JS.
```
Route (pages)                              Size     First Load JS
┌ ● /                                      922 B           101 kB
├   /_app                                  0 B            97.5 kB
├ ○ /404                                   182 B          97.7 kB
└ ● /second-page                           600 B           100 kB
+ First Load JS shared by all              97.5 kB
  ├ chunks/framework-7751730b10fa0f74.js   45.5 kB
  ├ chunks/main-c0428f9d2a1537c5.js        31.9 kB
  ├ chunks/pages/_app-eb818586830b4865.js  19.4 kB
  └ chunks/webpack-8fa1640cc84ba8fe.js     750 B

○  (Static)  automatically rendered as static HTML (uses no initial props)
●  (SSG)     automatically generated as static HTML + JSON (uses getStaticProps)
```

First load JS with the translation library seems to be around 100 kB. But do 100 kB nowadays really still matter? Every small image has more no? The .js of google.ch is also around 100 kB large. 

Doesn't need to be nextjs at all btw. I think everything, where we don't limit ourselves to purely static content and where we can add dynamic components' with ease (e.g. to automatically display numbers in the user's currency, to build the payment and admin interface) works for me.

> I just saw something similar in another project I started to work in. They use `react-intl` and [Accent](https://nextjs.org/learn/basics/data-fetching/with-data) to achieve something similar. To make it work in Storybook I just added the react provider wrapper of react-intl as a global story decorator in the Storybook config. It's mostly the same as your solution (except for the external translation editor), so I find that solution you are proposing great.

Yes, I also so that library. It seems that to me that they offer almost the same functionality. I don't have any preferences there. I just took i18next] because it had a bit more weekly downloads on npm. 😆 . 
react-intl (or format.js now?) also supports the [extraction of messages](https://formatjs.io/docs/getting-started/message-extraction/). If you already worked in another project with it, I would go with that one 👍

renestalder:
@andrashee

> First load JS with the translation library seems to be around 100 kB. But do 100 kB nowadays really still matter? Every small image has more no?

- It's not "a lot" in terms of what the average website is, but it is far too high for two blank pages which wouldn't need the client-side scripts anymore after static generation (they are redundant on the client). Like, I just checked what size the non-gzipped JavaScript bundle of one of my progressively enhanced (works also without JS) websites (multilingual) is, and it's around 100 kb non-gzipped and around 33 KB gzipped. That might give some dimensions.
- Somewhere between 140 kb and 200 kb is the recommended maximum for a JavaScript bundles. I can't remember the exact maximum number. But I know that somewhere around there you get your first warning in your bundler.
- 100 KB isn't much on a good 4G connection. It starts to be much on a 3G connection. It also maybe isn't that much for secondary elements like images, which have an alternative text and can be lazy loaded. The problem here is that the whole application hydrates at some point and if this happens only after like 10 seconds on a slow connection and the user is already interacting with the website, that can lead to problems. Usually, the web vitals measurements of Google Lighthouse will show those problems quite well.
- I would consider people from Sierra Leone might also access the website for information about the project and that would mean, we have to keep the footprint small enough or our optimization efforts big enough that the website runs well on slow networks.

If we can make next to generate it fully statically without hydrating the redundant parts again on the client, then that's fine. Otherwise, and I just wholeheartedly recommend it to my best knowledge and not because I prefer one tool over another, we should test with a couple of full static-side generators.

ps. I wouldn't necessarily say $BIG_CORP are good examples for "how it should be done" because often it's not the case 😅. That's also why there are [alternative versions of the google search engine for slow connections](https://play.google.com/store/apps/details?id=com.google.android.apps.searchlite&gl=US), which probably wouldn't be necessary if the process would be more inclusive from the beginning. Or like how the Microsoft website had a couple of 100 HTML validation issues for years, and Webkit shipping with workarounds to make those websites work in browsers like Safari.

> if you already worked in another project with it, I would go with that one

It doesn't matter for me, actually. I have nearly zero experience on that matter. I clearly enjoyed the developer experience of it, but I feel like your proposal here is mostly the same, and that's what I like about it. The approach matters, the tool not so much. :)

ssandino:
If the goal is to have a light website for specifically slow connections, then we could save much more with adding in the future a tool like [imagekit.io](imagekit.io) that can deliver images via CDN and based on network speed (as well as optimised for the used browser and if someone has low-data mode activated on a given device). 
 
For example: the animated house on the start page is 1mb. So if we would have a non animated version for slow connections, you could save almost the entire 1mb (as it could be delivered as svg and only has a few colors). 

Or even without such tool, we could just make a country version, where we take out the heaviest images and change them for lighter ones. In short: I see different solutions (based on content) to reach the goal of a light website in countries with slow connection. 

For the moment we have a very contributor-focused website. Which makes sense for now, as we also don't allow potential recipients to sign up for a Social Income (they are proposed via local NGOs). As contributors come from all over the (western) world it helps if we show some dynamic content (i.e. currency, language, but also in the future direct comparisons btw. IP-country and SL or showing first example companies (with contributors) that are from the IP-country etc.).

andrashee:
> Nice setup! Works well locally.
> 
> **2 Thoughts on Search Engine & Language** (1) I like that Next.js automatically adds the lang attribute to the `<html>`. This also helps to be better displayed on search engines. Now it's a bit messy in the search results as different languages are mixed up. We would then only have to add the additional `hreflang meta tags`.
> 
> Currently (Issue: it mixes languages, I thought Google is smarter than this): Screen Shot 2022-10-20 at 17 41 52
> 
> Nothing that couldn't achieved with some effort on the current website, but nice to have it out of the box.

The multilanguage sitemap should work out of the box. Added it to the demo.

```
<?xml version="1.0" encoding="UTF-8"?>
<urlset xmlns="https://www.sitemaps.org/schemas/sitemap/0.9"
        xmlns:news="https://www.google.com/schemas/sitemap-news/0.9" xmlns:xhtml="https://www.w3.org/1999/xhtml"
        xmlns:mobile="https://www.google.com/schemas/sitemap-mobile/1.0"
        xmlns:image="https://www.google.com/schemas/sitemap-image/1.1"
        xmlns:video="https://www.google.com/schemas/sitemap-video/1.1">
    <url>
        <loc>https://socialincome.org</loc>
        <lastmod>2022-10-21T09:36:41.890Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org"/>
    </url>
    <url>
        <loc>https://socialincome.org/second-page</loc>
        <lastmod>2022-10-21T09:36:41.890Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org/second-page"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org/second-page"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org/second-page"/>
    </url>
    <url>
        <loc>https://socialincome.org/de</loc>
        <lastmod>2022-10-21T09:36:41.891Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org/de"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org/de"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org/de"/>
    </url>
    <url>
        <loc>https://socialincome.org/de-CH</loc>
        <lastmod>2022-10-21T09:36:41.891Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org/de-CH"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org/de-CH"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org/de-CH"/>
    </url>
    <url>
        <loc>https://socialincome.org/de/second-page</loc>
        <lastmod>2022-10-21T09:36:41.891Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org/de/second-page"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org/de/second-page"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org/de/second-page"/>
    </url>
    <url>
        <loc>https://socialincome.org/de-CH/second-page</loc>
        <lastmod>2022-10-21T09:36:41.891Z</lastmod>
        <changefreq>daily</changefreq>
        <priority>0.7</priority>
        <xhtml:link rel="alternate" hreflang="en" href="https://socialincome.org/de-CH/second-page"/>
        <xhtml:link rel="alternate" hreflang="de" href="https://socialincome.org/de-CH/second-page"/>
        <xhtml:link rel="alternate" hreflang="de-CH" href="https://socialincome.org/de-CH/second-page"/>
    </url>
</urlset>

```

Alternatively, according to Google, one can also just add the alternate links into the head part of the individual site instead of adding a sitemap: https://developers.google.com/search/docs/specialty/international/localized-versions

> **1 Thought on Translations** The workflow makes sense to me. As we have more languages, we'll always end up in the situation, where we have a new page or element ready to deploy, but not yet in all languages. And to be not hold back by translations (which can take time, when we are depending on outside people) it would be nice if we can also deploy with a missing language. In this case, it should just fall back or display the default (EN) until the missing translation keys are added. Is this possible @andrashee ?

Yes, it seems that one can set a setting `returnEmptyString: false` to fallback to the default language in case a translation is missing.

andrashee:
> @andrashee
> If we can make next to generate it fully statically without hydrating the redundant parts again on the client, then that's fine. Otherwise, and I just wholeheartedly recommend it to my best knowledge and not because I prefer one tool over another, we should test with a couple of full static-side generators.

For me, it doesn't really matter what framework we are going to use. I personally would not weight the slightly bigger js file as high of a priority if that's the main argument against a tool. Especially when we are not optimizing everything else, but using custom fonts and so on.

Due to the lack of experience with web development, I just have a hard time understanding how we can build the payment process / admin part with full static-side generators. How would we do that?

> 
> ps. I wouldn't necessarily say $BIG_CORP are good examples for "how it should be done" because often it's not the case 😅. That's also why there are [alternative versions of the google search engine for slow connections](https://play.google.com/store/apps/details?id=com.google.android.apps.searchlite&gl=US), which probably wouldn't be necessary if the process would be more inclusive from the beginning. Or like how the Microsoft website had a couple of 100 HTML validation issues for years, and Webkit shipping with workarounds to make those websites work in browsers like Safari.
> 

Absolutely agree, that they are not necessarily good examples “how it should be done”, but at the same time it does IMHO put the “how it theoretically should be done” into perspective how important it is in practice.

andrashee:
> If the goal is to have a light website for specifically slow connections, then we could save much more with adding in the future a tool like [imagekit.io](imagekit.io) that can deliver images via CDN and based on network speed (as well as optimised for the used browser and if someone has low-data mode activated on a given device).
> 
> For example: the animated house on the start page is 1mb. So if we would have a non animated version for slow connections, you could save almost the entire 1mb (as it could be delivered as svg and only has a few colors).
> 

I agree. Many frameworks have now quite nice integrations for image & font optimizations:
- https://nextjs.org/docs/basic-features/image-optimization
- https://nextjs.org/docs/basic-features/font-optimization

Not sure how this compares to imagekit in practice, but could be a start.

renestalder:
>  I just have a hard time understanding how we can build the payment process / admin part with full static-side generators. How would we do that?

There are basically the 3 approaches I know:

- Everything is server-side rendered, like the current website, which is a full Node application (rendered on a serverless function, which is a bit of a contradiction to what those are made for). You can also do the same with NextJS, but will have the same problem: Even though only a few parts of the UI need some client-side scripts, it will hydrate the whole website back to the client.
- Everything is rendered as a JavaScript only Single-Page Application (SPA). Will not work without JavaScript activated, will fail when JavaScript has runtime errors, and the first page load will just take way longer. This is probably the worst trend that got traction for building websites.
- Everything is statically generated. Dynamic parts like a donation calculator are JavaScript rendered (Island architecture for example or just manually with separate JavaScript files). Dynamic parts that require backend business logic like the login, send their request to a serverless function, that handles the request and sends back the result.
  - The user area, which is completely individual to every user, cannot be statically generated and must either be server-side rendered via a serverless function (like it is now) or is rendered client-side (as an SPA).
  - Everything else that looks the same to every user can statically be generated ahead of time.

I completely understand why this is perplexing. The whole web industry turned full circle from 2008 to 2022, going from HTML pages to server-side only rendered websites to single page applications, to hybrid solutions like Next and now back HTML with progressively enhance individual parts of the website (and gave it a fancy name like islands architecture, even though, that's not really new).

renestalder:
> f the goal is to have a light website for specifically slow connections, then we could save much more with adding in the future a tool like [imagekit.io](https://github.com/socialincome-san/public/pull/imagekit.io) that can deliver images via CDN and based on network speed (as well as optimised for the used browser and if someone has low-data mode activated on a given device).

We don't even need a service for that. Either we at some point at a CMS, which will provide the functionality (like Storyblok does) or if we don't use a CMS but manually add content and images in the repository, we can do it like @andrashee says, with any built-in function of our tooling. Or we can also resize images in GitHub actions and upload them to Firebase storage.

ssandino:
> 👍 Thanks a lot for the detailed explanation. So would you go for either Astro or 11ty with Islands?

This one @renestalder

renestalder:
Yeah sorry I totally "overheard" the last part.

> So would you go for either Astro or 11ty with Islands?

As we now settled with React, definitely Astro or some other React based static-side solution with Islands support. I am not confident how far Gatsby is on that front (I only saw there is an experimental plugin to it), but I'd skip Gatsby from consideration due to other reasons that might not fit our vision.

Eleventy is more the tool for when you keep it bare bones.

andrashee:
In the meantime, I skimmed a bit through the documentation of Astro and played a bit around with it.

While I definitely see its advantages in purely static websites, I'm honestly still not convinced that it is the better solution for our project.

Here my reasoning:
I think many of our landing pages will involve react components like e.g. the contribution calculator, currency or language switcher. Most likely, we will have components with a `client:load` (or similar) directive to ship JS code to the client.

If you decide to use React for this, then the advantages of Astro are IMHO gone. 

Out of curiosity I built a super simple example of a counter, once with Next and once with Astro and simulated the behaviour with a slow 3G connection:

Astro:
image

image

Next:
image

image


Time till the page is visually loaded (FCP) is the same for both (4s). Which makes sense, since for for both the html is statically built. This is IMHO the important metric. 

Time to interactive (the time needed to load the JS and to hydrate it): 6s for next and 8s for Astro (somehow the js only gets loaded after the css, but I guess that's somehow configurable).

For Astro this can be improved by using e.g. preact instead of react to save some bytes. But in my opinion, this is rather irrelevant since the time to visually render the page will be very similar for both solutions.

My impression, however is, that we have more flexibility when we go with next. A few things are coming into my mind which I think I prefer with my current stand of knowledge:
- i18n works well for server side and client side rendering. While static generation seems to work also for Astro, I'm not fully sure how well the dynamic one already works: https://github.com/yassinedoghri/astro-i18next/issues/46.
- Next has the concept of [Incremental Static Regeneration](https://nextjs.org/docs/basic-features/data-fetching/incremental-static-regeneration) where you can say that the server should rebuild a page every x time interval. I imagine this a nice feature, e.g. for the transparency page. Instead, that an Astro client island needs to get the current stats from the database, the server can retrieve it periodically, pre-render and cache it. 
- Next feels more mature. Easier to find open source contributors.

renestalder:
@andrashee you are comparing the render blocking CSS in the Astro example with the one in the Next example. That's more an issue of our CSS there in the waterfall diagram. We clearly have an issue there with the CSS due to how Tailwind works and how I mistakenly added a too broad safelist in the config that makes the CSS unnecessary huge. That can be fixed.

I will check that next week.

I want that we make a decision focused on user experience, not on developer experience. I believe that the latter is a bad metric to create the former. If it turns out that Astro doesn't hold up to what they claim, then I have no issue with dropping that idea. Until then, I can't just blindly opt-in for the most popular solution while being aware that the output of this, what we, web developers that went through all these loops many times, [is not how websites should be](https://twitter.com/zachleat/status/1584995586918731776?s=46&t=9E3OSbB53zGdvF4kyehR7Q).

This just can be done with a smaller JavaScript overhead on the client. Otherwise our current website wouldn't work either.

ssandino:
I enjoy this conversation, as it brings up fundamental questions around how to decide on the right tools & technology (regardless of the choice we make in the end). It stands out that there will be also pros and cons to any solution, important is that we discuss them, let everyone weigh in and then choose on the end a solution.  

The motto _decision focused on user experience_ as guiding principle made me reflect on the overall user experience and the top 3 things that matter to me as a personal user:

1. **speed** (we tend to click away slow loading pages. I am positive, all solutions discussed above could reach a decent speed. speed improvements we can get from scaling down images, lazy loading etc. something to integrate in any future setup)
2. **design** (let's just say it shouldn't be repellent or outdated. Our design should exude a sense of trust and confidence. I think we already have a good base with the radical transparency approach. This is not really part of this discussion.)
3. **relevant information** (I mean here to see information that triggers my attention, information that holds an added value and information that somehow relates to "own lil world")

I'd like to highlight the third point. To order my thoughts I made below a quick sketch of how the landing page could change depending on what kind of user accesses the page. There are various differentiating factors such as:

- location (effects currency shown, testimonials and logos shown etc.)
- language 
- logged in status (effects what kind of additional options are highlighted for user)
- first visit, or revisit after filling out calculator (important: this should never be saved if we have no cookie consent)
- device visiting (if with a mobile phone the story changes a bit -> _mobile to mobile_!)

This should help to assess if Astro can handle such dynamic parts in a smart way (individual or in any combination) – this being the question that seems to be unanswered.  Furthermore, it seems not yet clear to me how the contributor dashboard (behind the login) would be made with Astro.

We are not yet using the full potential of personalisation, however, this is where the journey is going (aka creating a compelling customer experiences). More contributors allow to support more recipients (this also being a guiding principle of Social Income and the idea of redistribution of wealth). 

@mkue and @brennerthomas your opinions also always welcome, especially as you have been working with many tools and setups in the past. 

3Artboard 9-100

mkue:
> I want that we make a decision focused on user experience, not on developer experience. I believe that the latter is a bad metric to create the former. If it turns out that Astro doesn't hold up to what they claim, then I have no issue with dropping that idea. Until then, I can't just blindly opt-in for the most popular solution while being aware that the output of this, what we, web developers that went through all these loops many times, [is not how websites should be](https://twitter.com/zachleat/status/1584995586918731776?s=46&t=9E3OSbB53zGdvF4kyehR7Q).

I agree with this, but I kind of fail to see how using Next.js would not provide a performant user experience. As Andras has pointed out, time to render (which I believe is the most important metric) will be very similar no matter if we use Astro or Next.js. Also, I just deployed a Next.js website at work, and it had a 100% performance score in the Lighthouse report in Google Chrome. In my opinion, we are trying to preemptively optimize something that does not seem to be an issue (correct me if I'm wrong🙈). 

Another aspect we should not forget is that we all work on this project only in our free time and therefore have limited time. For this reason, we should use technologies that most of us already know. This way, we will be able to rebuild and expand the website as fast as possible. Therefore, I would suggest that we definitely use React together with either Next.js or Astro. Personally, I would prefer Next.js because I already know it a little bit, but I have no strong opinion here.

One last question that I have: As far as I understand it, the main downside of Next.js is that it requires a 100kb sized JS bundle. Isn't 100kb extremely small? If those 100kb are a problem, shouldn't we also rigorously optimize every other asset that we serve? Fore example, the firestore bundle is 336kb and the fonts we use are together also 300kb. Why is that not a problem as well?

andrashee:
> @andrashee you are comparing the render blocking CSS in the Astro example with the one in the Next example. That's more an issue of our CSS there in the waterfall diagram. We clearly have an issue there with the CSS due to how Tailwind works and how I mistakenly added a too broad safelist in the config that makes the CSS unnecessary huge. That can be fixed.

This is not related to any of your changes. In this mini demo, I'm not including your UI or tailwind, but just added one simple plain css class.
Maybe one can configure Astro somehow to already start with the JS download in parallel instead of first loading the css. I haven't looked into this. 

> I want that we make a decision focused on user experience, not on developer experience.

I think we are all aligned that we want to optimize on the user experience. I just massively doubt that a few JS bytes are a significant contributor to this one. For me, user and developer experience go hand in hand. If one can develop much more features in a shorter time in comparison to another, it might lead to a better user experience. E.g. In my opinion, having just one framework for both the website and administration part for the user will lead to a better user experience because we (part-time) developers will be just quicker in implementing features, rather than spending the time in learning two frameworks.

To me, Astro seems like an awesome product for purely static pages. But I think we don't have just those, and at some point our users will need to load the hydration framework anyhow.

- Our landing page will have dynamic content like the calculator. I think we all agree that we want to use React for those, right? If we use `client:load` then Astro transferred actually more js than nextjs in my example. Maybe I did something wrong, but once you want a React island, the differences seem to be small.
- How would you do the transparency page https://socialincome.org/ch/en/finances which should get updated frequently (let's say every 10 minutes) with Astro?
- Would the administration part also be islands?

>If it turns out that Astro doesn't hold up to what they claim, then I have no issue with dropping that idea. Until then, I can't just blindly opt-in for the most popular solution while being aware that the output of this, what we, web developers that went through all these loops many times, [is not how websites should be](https://twitter.com/zachleat/status/1584995586918731776?s=46&t=9E3OSbB53zGdvF4kyehR7Q).

What pieces are missing there to make a decision?

renestalder:
> One last question that I have: As far as I understand it, the main downside of Next.js is that it requires a 100kb sized JS bundle. Isn't 100kb extremely small? If those 100kb are a problem, shouldn't we also rigorously optimize every other asset that we serve? Fore example, the firestore bundle is 336kb and the fonts we use are together also 300kb. Why is that not a problem as well?

Because we are talking about a 100 Kb baseline of JavaScript that is obsolete. That is the foundation of an application that is hydrated on the client, that wouldn't be necessary for the site to run. It just makes it twice: Once on the server in the SSG task and then again on the client (for no apparent reason other than running the site on the client after the first load).

The Firestore bundle can only be loaded for the login, right? That's for the user database. There is no need to load it for the whole website the whole time.

Fonts can't be "server-side" rendered and for sure, 300 KB of web fonts is outrageous.

renestalder:
Okay, look folks, I accept that you all want to go with Next. I see many open questions about how certain things will be solved with a statically generated websites. It's not like this has never been done before. I could give you dozens of example, e.g. https://www.netlify.com/ is statically generated. https://www.smashingmagazine.com is statically generated. Both websites are way more complex than ours.

I don't want to stretch this conversation unnecessarily. I have shared my experience as good as I can. It's better to move forward now.

Therefore, as the major consent is to use Next, I guess that's settled, and I propose that somebody takes the responsibility for guaranteeing performance, SEO and accessibility of this new website, that the information on the website can still be retrieved with JavaScript broken/disabled and that it can beat our current server-side rendered website, that loads client-side scripts for individual components. Because that is the actual metric from my point of view. I would like to see that somebody takes over the lead. Somebody who feels well at home with Next.

I'm happy to help with building the new website with whatever you want to use, but as you say, we all volunteer, and the only thing I'm not going to donate my time for is trying to fix bottlenecks in a growing JavaScript application, whose framework might shut me out of doing so. I've been there many times, with Angular (as a whole), with Vue + Nuxt, it's a hell of an effort to compete SSG focused tooling or no tooling at all, I'm not going into that again.

ssandino:
> Web fonts can't be just "server-side" rendered and definitely, 300 KB of web fonts is outrageous. 

Yes, I think in the new setup we can get down to 90 KB in total. This is my secret plan: 

1. We scrap the `fa fonts` on the new website. This [Font Awesome](https://fontawesome.com) fonts are for the icons we use (as text) and came with the Bootstrap theme Front. Given that we drop the Front theme and anyway not using lots of the icons provided by the Font Awesome, we can just use individual small SVGs for the existing icons we use currently.
2. We can load only a subset of the Unica77 fonts (Bold, Regular we need), which are each 44 KB instead of heavy full fonts, which weigh about 75 KB each. (_Subsetting is the practice of creating a “subset” of a font – a file that contains a custom (and usually limited) collection of glyphs_.)

**Task to myself:** open up an issue to upload the subset for the new website. 

What we load at the moment:
Screenshot 2022-11-14 at 17 32 02

Thanks for this discussion above.

ssandino:
Nice, Vercel gives us also the region `(x-vercel-ip-country-region),` which could allow for future individualisation of specific content (e.g. "Zurich's average salary is nXmore compared to Freetown). Not sure what naming convention they follow. 
https://edge-functions-geolocation.vercel.sh

ssandino:
@andrashee could you sneak in an empty page with the URL .../social-responsibility in the next update. This allows @soumdatta81 to start with #181.

andrashee:
> @andrashee could you sneak in an empty page with the URL .../social-responsibility in the next update. This allows @soumdatta81 to start with #181.

I can add this page in the stacked PRs with the proper setup. I will close this one since it was only for demo purposes.

PR #144, opened 2022-10-21
[functions] Stripe webhook to ingest charges into firestore

andrashee:
Still WiP. What is still missing:

- [x] configuring the secrets in github
- [ ] reconfigure the webhook on the stripe webpage once this is merged
- [x] creating a script to add all the past payments


ssandino (approved):

ssandino:
Quick question: Does it import also information about failed payments or only successful payments?

andrashee:
> Quick question: Does it import also information about failed payments or only successful payments?

Good point! Changed the logic to also handle and ingest failed ones.

PR #148, opened 2022-10-26
feature: integrate firestore storage rules

mkue:


mkue:
Oh wow, this took me forever🙈 Renaming the project from `demo-social-income-prod` to `social-income-prod` lead to some weird side-effects with the tests because the document references we use in some firestore collections, e.g. recipient.organisation, pointed to documents in the emulated `demo-social-income-prod` firestore. Unfortunately, you never see the full document path with the project id included, so it wasn't obvious why some document queries in the tests just didn't work...

andrashee:
Very nice @mkue, I will have a look at it tomorrow!

.github/workflows/firestore-rules.yml
```diff
@@ -0,0 +1,33 @@
...
+    steps:
+      - name: Checkout Repo
+        uses: actions/checkout@v2
+
+      - name: Install firebase-tools
+        run: npm -g install firebase-tools
+
+      - name: Deploy to Firebase
+        if: github.ref == 'refs/heads/main'
+        run: firebase deploy --only storage --token ${{ secrets.FIREBASE_TOKEN }} --project social-income-prod
```
andrashee, line 33: is this correct or should it be `--only firestore:rules`?

firebase.json
```diff
@@ -29,7 +29,7 @@
 		"runtime": "nodejs16"
 	},
 	"storage": {
-		"rules": "storage.rules"
+		"rules": "firestore.rules"
```
andrashee, line 32: Is this change on purpose? Here we are talking about storage rules not firestore no?

functions/package.json
```diff
@@ -9,9 +9,8 @@
 		"build": "tsc",
 		"build:watch": "tsc --watch",
 		"serve": "npm run build:watch",
-		"serve:watch": "npm run build -- --watch & firebase emulators:start --project social-income-prod --config ../firebase.json --import ../seed",
-		"test": "firebase emulators:exec --only functions --only firestore --project social-income-prod --config ../firebase.json 'npm run test:app'",
-		"test:app": "jest --roots  src/"
+		"test": "firebase emulators:exec --only functions --only firestore --project social-income-prod --config ../firebase.json --import ../seed 'npm run test:app'",
```
andrashee, line 12: Why did you delete the serve:watch? I find it useful to work with.

recipients_app/.run/prod.run.xml
```diff
@@ -1,7 +0,0 @@
-<component name="ProjectRunConfigurationManager">
```
andrashee, line 1: Do you delete this on purpose?

recipients_app/.run/dev.run.xml
```diff
@@ -1,7 +0,0 @@
-<component name="ProjectRunConfigurationManager">
```
andrashee, line 1: Do you delete this on purpose?

storage.rules
```diff
@@ -1,8 +0,0 @@
-rules_version = '2';
```
andrashee, line 1: Why did you delete this one?

andrashee (commented):
Very nice 👍 

Is the current file up-to-date with the in firebase?

.github/workflows/firestore-rules.yml
mkue: Oh wow, you're absolutely right. This is completely wrong!

I changed it, we deploy both rules files now.

firebase.json
mkue: Yes, you're right. I totally mixed this up.

recipients_app/.run/prod.run.xml
mkue: Hmm, I think I removed some run configurations in WebStorm that were suddenly there. I've never shared IDE run configurations in a repository and I don't think we should. What do you think?

recipients_app/.run/dev.run.xml
mkue: Same as above

storage.rules
mkue: Because I made a mess😅

functions/package.json
mkue: Oh sorry, this is a mistake. I thought that it can't be executed, which is totally wrong.

mkue:
@andrashee, please quickly take a look again.

> Is the current file up-to-date with the in firebase?
Yes it is👍🏼

andrashee (approved):
That's awesome! I find it a massive step to have those rules manageable and testable. 🥳

ssandino:
🎉🎉🎉

PR #152, opened 2022-11-05
[admin] Scripts view to trigger backend functions

andrashee:
A custom page (for global admins only) to trigger non-scheduled backend jobs. Could replace most of the current scripts.

image


ssandino:
Nice! @sdwivedi wouldn't that be an easy way to do run the monthly script to set the payment status of all recipients from `to be paid` to `tobe confirmed`.

Easier than it a button in the collection view. Could be a start to have it here, as it's anyway run by admins only.

PR #155, opened 2022-11-06
feat(admin/functions): initial version of donation certificates

brennerthomas:
- Adding new type Donation Certificate
- Adding new subcollection for User
- Adding functionality to create PDF for Switzerladn
- Adding uploading capabilities to storage


functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+
+export const bulkDonationCertificateBuilderFunction = functions.https.onCall(async (data, context) => {
+	const app = await initializeFirebaseApp();
+	const firestore = await initializeFirestore(app);
+	const now = new Date();
+	const currentDate = now.toLocaleDateString('de-DE');
+
+	const calculationYear = data.year
+
+	for (let i = 0; i < data.users.length; ++i) {
```
andrashee, line 19: shall we use a for loop?
```
for await (const user of data.users ) {
}
```

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+import { initializeApp } from "firebase/app";
+import { getFirestore, connectFirestoreEmulator } from 'firebase/firestore';
+import { doc, setDoc, getDocs, collection } from "firebase/firestore";
+import { createAndUploadDonationCertificate } from './generateDonationCertificatePDF';
+import * as config from '../config';
+
+let FIRESTORE_EMULATOR_REQUIRED = true;
+
+export const bulkDonationCertificateBuilderFunction = functions.https.onCall(async (data, context) => {
+	const app = await initializeFirebaseApp();
```
andrashee, line 12: I think you should be able to reuse the `firestore` singleton in the `useFirestoreAdmin.ts`:
https://github.com/socialincome-san/public/blob/614605df94c297aa62b56d6d53fe9a036e576067/functions/src/useFirestoreAdmin.ts#L16

Then you can remove the `initializeFirestore` and `initializeFirebaseApp` here.

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+		}
+		console.log('Using firestore emulator');
+	} else {
+		console.log('Using production firestore');
+	}
+
+	return firestore;
+}
+
+export const prepareData = async (data: any, calculationYear: string, currentDate: string, firestore: any) => {
```
andrashee, line 62: I would use the proper types instead of `any`, then we can profit more from type checking.

shared/types/admin/DonationCertificate.ts
```diff
@@ -0,0 +1,8 @@
+export const DONATION_CERTIFICATE_FIRESTORE_PATH = 'donation-certificate';
+
+export type DonationCertificate = {
+	created: Date;
+	url: string;
+	country: string;
+	year: string;
```
andrashee, line 7: nit: number

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+
+
+	const donationCertificateRef = collection(firestore,"users", data.id, "contributions");
+
+	const querySnapshot = await getDocs(donationCertificateRef);
+
+	let contributions: any[] = [];
```
andrashee, line 69: Shall we use a `.map` here? 

```
const contributions = querySnapshot.map(element => element.data())
```

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+		currentDate: currentDate,
+		address: data.values.address,
+		personal: data.values.personal,
+		financials: await calculateFinancials(contributions, calculationYear)
+	}
+
+	return pdfData;
+}
+
+export const calculateFinancials = async(contributions: any[], year: string) => {
```
andrashee, line 87: types: `contributions: Contribution[], year: number`

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
```diff
@@ -0,0 +1,140 @@
...
+	let total_chf = 0;
+	let total_usd = 0;
+	let total_eur = 0;
```
andrashee, line 90: Maybe a bit of an overkill at this point, but what do you think of making this dynamic instead of hardcoding the currencies?

You can use lodash for nicer functional operations on js collections.
```
npm install  lodash
```

And then something like 
```
import _ from "lodash";
const totalContributionsByCurrency = _(contributions)
	.filter(contribution => contribution.created?.toDate()?.getFullYear() === year)
	.groupBy(contribution => contribution.currency)
	.mapValues(contributions => _.sumBy(contributions, contribution => contribution.amount))
```

Which gives you then a `currency -> totalAmount` pair.

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
```diff
@@ -0,0 +1,156 @@
...
+	translations.set('donation-certificate.thanks','Herzlichen Dank für Ihr Vertrauen und Engagement.');
+	translations.set('donation-certificate.greetings','Freundliche Grüsse,');
+	translations.set('donation-certificate.sender',`Sandino Scheidegger\nCEO / Founder\nss@socialincome.org`);
+	translations.set('donation-certificate.single-donations', 'Liste Einzelspenden')
+	return translations;
+
+}
+
+
+export const createDefaultDonationCertificate = async () => {
```
andrashee, line 154: nit: delete?

shared/types/admin/DonationCertificate.ts
```diff
@@ -0,0 +1,8 @@
+export const DONATION_CERTIFICATE_FIRESTORE_PATH = 'donation-certificate';
+
+export type DonationCertificate = {
```
andrashee, line 3: We need to make sure that we set the correct firestore rules for this collection, so one can't see the `DonationCertificate` entries of others.

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
```diff
@@ -0,0 +1,156 @@
...
+		uploadFileAndCreateDocumentEntry(fileRef, pdfData, data, firestore, storage);
+	});
+
+	docPdf.end();	
+}
+
+
+export const addDocumentToUser = async(data: any, fileRef: any, firestore: any, storage: any) => {
+
+	const downloadUrl = '[File Download](' +  await getDownloadURL(fileRef) + ')';
```
andrashee, line 116: Have you already put some thoughts into how we ensure that one can only download the certificates of oneself? 
Not that one can get donation information of others by reverse engineer the urls.

storage.rules
```diff
@@ -2,7 +2,8 @@ rules_version = '2';
 service firebase.storage {
   match /b/{bucket}/o {
     match /{allPaths=**} {
-      allow read, write: if false;
+      // Allow access by all users
+      allow read, write;
```
andrashee, line 6: That sounds a bit dangerous 😆 We have several rules defined for firestore. Let's sync on how we can ensure that only a user (and a global admin) can only download his/her certificate.

andrashee (commented):
Very nice @brennerthomas! This will such an admin time safer. 👍 

Some high-level comments for now. I think the most important one for me is that we make sure that we have proper rules for the PDF files and db entries.

side note: curious why the github actions did not run for this PR. I assume it's because it's your first PR? Will have a look.

admin/src/collection-actions/bulkDonationCertificiateCreation.tsx
```diff
@@ -0,0 +1,84 @@
...
+                })
+                .catch((err) => {
+                    console.log(err);
+                    snackbarController.open({
+                        type: "error",
+                        message: `An error occured during certificate creation.`
+                    });
+                });
+        } else {
+            snackbarController.open({
```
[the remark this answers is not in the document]
ssandino: Ah nice to add this snackbar!

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
ssandino: Could this be also solved with the approach we discussed around the survey links? -> Using custom tokens?

[Create custom tokens using the Firebase Admin SDK](https://firebase.google.com/docs/auth/admin/create-custom-tokens#create_custom_tokens_using_the_firebase_admin_sdk)

> We can create the token when generating the survey links. The token contains the recipient's uuid. The react app can then call the auth endpoint firebase.auth().signInWithCustomToken(token). Once authenticated it can query the survey collection. In the request, the token is passed along and one can read the recipient's uuid in the security rules. I would go for this approach since jwt token authentication is a standard and widely used.

ssandino:
Thanks 🙏 this will be a game changer

admin/src/App.tsx
```diff
@@ -132,7 +133,12 @@ export default function App() {
 			firebaseConfig={config.FIREBASE_CONFIG}
 			onFirebaseInit={onFirebaseInit}
 			dateTimeFormat={'yyyy-MM-dd'}
-			toolbarExtraWidget={<CallDummyFunctionButton />}
+			toolbarExtraWidget={
+				<div>
+					<CallDummyFunctionButton />
+					<CallPDFBuilderFunction />
```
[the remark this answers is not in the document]
brennerthomas: I saw that one, I thought that this functionality is too "specific" as a common action, so I moved it into the "Users" view.

admin/src/collections/Users.ts
```diff
@@ -177,7 +181,14 @@ export const usersCollection = buildCollection<User>({
 		},
 		location: {
 			name: 'Location',
+			description: "Living location defined by List of ISO 3166 country codes",
 			dataType: 'string',
+			defaultValue: 'CH',
+			validation: {
```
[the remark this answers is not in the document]
brennerthomas: Thanks, I will mark it as resolved in here

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
brennerthomas: Already changed. @andrashee, I would be interested why this is better than for loop

admin/src/App.tsx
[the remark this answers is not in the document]
andrashee: That makes total sense 👍 . Do we still need the toolbar widget then?

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
andrashee: In this context, I think it doesn't really matter and is a more cosmetic question which one to prefer. I find the syntax of `for each` the more modern and appealing way, but feel free to use whatever you like more.

Generally, I guess one tries to abstract away implementation details (e.g. in this case how to access elements in a data structure) as much as possible. This allows easier modularization and reuse of methods.

Your implementation assumes that `data.users` is an indexed data structure (e.g. array), but for the use case it actually only needs to be iterable. E.g. if you have a function like:
```
const createAndUploadDonationCertificate = async (user: User) => {{}}

const createAndUploadDonationCertificates = async (users: Iterable<User>) => {
	for await (const user of users) {
		await createAndUploadDonationCertificate(user);
	}
}
```

Then you can feed it with whatever data structure implements the iterable interface, and you delegate the implementation details of looping to the data structure. E.g. index based for arrays as your implementation, pointers to the next entries for linked lists and streams, ... 
For example, if we all of a sudden would have many users, one could then easily switch to read a user stream from firestore instead of loading the entire array into memory to keep the memory pressure O(1) instead of O(users). 

But this more as a high-level explanation, not that it matters in our case here :-).

admin/src/CallPDFBuilderFunction.tsx
```diff
@@ -0,0 +1,26 @@
+import { PictureAsPdf } from '@mui/icons-material';
+import { IconButton, Tooltip } from '@mui/material';
+import { getFunctions, httpsCallable } from 'firebase/functions';
+
+export default function CallPDFBuilderFunction() {
+	const onClick = () => {
+		const functions = getFunctions();
+		console.log(functions);
+		const pdfBuilderFunction = httpsCallable(functions, 'pdfBuilderFunction');
+		pdfBuilderFunction(55)
```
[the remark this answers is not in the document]
brennerthomas: Yes, deleted

admin/src/App.tsx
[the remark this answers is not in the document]
brennerthomas: not for me :-)

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
brennerthomas: Could use the Singleton for Firestore, for the Storage it was not possible for me. Perhaps we can solve that in a second step :-)

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
brennerthomas: Here I am not sure how to deal with JSON, if you could provide some help, it would be very much appreciated. I adapted in all the functions the types where I was sure :-)

shared/types/admin/DonationCertificate.ts
brennerthomas: I actually used string on purpose because I use it only as a string

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
brennerthomas: querySnapshot has no property `map`

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
brennerthomas: The donation certificates use the ID of the user in the url and I attached an additional random string at the end to make sure reverse engineering is not possible.

storage.rules
brennerthomas: Yes :-) Can you help me with that? Here I am in deep waters @andrashee 🤣

andrashee: I'm also not an expert by all means. My approach would have been:
- serve the `donation-certificates` publicly. I think for our use case, where we just want to send someone the URL retrieved through `getDownloadURL` per email and where that link is directly downloadable (without going through a login), those paths need to be publicly readable.
- to avoid any unwanted leakage in the future, I would leave the remaining paths private?

The admin storage api (when it's execute within a firebase function, the private key is automatically populated) can circumvent those roles. For that, it's important that you load it like this:

```
const { getStorage } = require('firebase-admin/storage');
```

See https://firebase.google.com/docs/storage/admin/start

We do the same for the firestore part. Frontend parts (e.g. the admin react interface or the website) are going through the public sdk and need to follow the security rules. In the functions we are using the admin sdk which can do everything.

So something like this (I don't know the exact syntax)?
```
service firebase.storage {
   match /b/{bucket}/o {
     /donation-certificates/{multi_path=**} {
         allow read;
     }
     match /{allPaths=**} {
       allow read, write: if false;
     }
   }
 }
```

@mkue wdyt?

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
```diff
@@ -0,0 +1,156 @@
+import { connectStorageEmulator} from 'firebase/storage';
+import { getStorage, ref, uploadBytes, getDownloadURL } from "firebase/storage";
+import { doc, setDoc } from "firebase/firestore";
+import * as config from '../config';
+
+let STORAGE_EMULATOR_REQUIRED = true;
+
+export const createAndUploadDonationCertificate = async (data: any, app:any, firestore: any) => {
+
+	const storage = getStorage(app);
```
andrashee, line 10: I would suggest that we have a separate file `useStorageAdmin.ts` which exposes the admin storage (see explanation between the difference in the security rule part). Then we have a consistent setup for storage and firestore. Let me create a separate PR for this 👍

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
andrashee: True, external data is always tricky.
I think you could create an interface for it. Then one can see in one glance what we expect from the request.

```
interface Data {
 myexpectecedVariable: string
}
```

functions/src/pdfBuilder/donationCertificateBuilderFunction.ts
andrashee: ah might be `querySnapshot.docs.map(element => element.data())`

functions/src/pdfBuilder/generateDonationCertificatePDF.ts
andrashee: https://github.com/socialincome-san/public/pull/161

storage.rules
andrashee: Update: when we use the admin storage linked here https://github.com/socialincome-san/public/pull/161, we can leave this rules to `allow read, write: if false;`. I think this is much better, otherwise everyone could have written to our public storage.

mkue (approved):
Very very cool stuff @brennerthomas! Impressive first contribution👏🏼👏🏼👏🏼 As we discussed, I cleaned up the code a bit and improved the typing. See #184 for all my changes.

PR #156, opened 2022-11-06
[functions] parallel stripe batch ingestion

andrashee:
Ingesting the stripe charges into firestore sequentially times out after 540 secs which is the max runtime for a firebase v1 function. Latency from our firebase functions to firestore is surprisingly slow. Could be because the functions are running in the US, but our firestore is in Europe? At some point, we should run the functions in the same DC.

Switching to parallel ingestion.

### DIFF
--- a/functions/src/etl/stripeWebhook.ts
+++ b/functions/src/etl/stripeWebhook.ts
@@ -114,6 +114,7 @@ export const batchImportStripeChargesFunc = functions
 	})
 	.https.onCall(async () => {
 		const stripeBatchSize = 100; // max batch size supported by stripe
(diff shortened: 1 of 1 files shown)


ssandino:
> Could be because the functions are running in the US, but our firestore is in Europe? At some point, we should run the functions in the same DC.

There is the option to change ([Best practices for changing region](https://firebase.google.com/docs/functions/locations#best_practices_for_changing_region)), BUT not sure if an exception applies to us: _"If you are using HTTP functions to serve dynamic content for Firebase Hosting, you must use us-central1."_

PR #160, opened 2022-11-12
[functions] create user when importing stripe charges

andrashee:
At the moment, in the stripe webhook, we dismiss stripe charges which we can't assign to a user.
I implemented it in that way, assuming, that our website always creates a user. However, this is not the case when the initial payment failed. 

Since in a next step, we would like to send custom emails to customers for which the payments fail, I propose here to create a barebone user from the information we have in the stripe charge.


ssandino (approved):

shared/utils.ts
```diff
@@ -1,3 +1,15 @@
+const capitalizeString = (value: string) => {
```
mkue, line 1: Should we install lodash and use `_.upperFirst()`?

Lodash will be useful anyway, I think

https://lodash.com/docs/4.17.15#upperFirst

mkue (approved):
nice, lgtm👍

shared/utils.ts
andrashee: Much better indeed, thanks!

shared/utils.ts
```diff
@@ -1,3 +1,13 @@
+import _ from 'lodash';
+
+export const capitalizeStringIfUppercase = (value: string) => {
+	if (value.toUpperCase() === value) {
+		return _.upperFirst();
```
mkue, line 5: nit: You need to pass the value to the function😉

andrashee: hahah, that's indeed beneficial. thanks for rechecking. good that all my tests capture it 😆

PR #161, opened 2022-11-12
[functions] Storage Admin

andrashee:
I would like to have a StorageAdmin helper class analogue to the Firestore one.

Doing so, I wanted to add a test to ensure that the storage rules work as expected.
I'm uploading a file using the admin sdk, and then I'm retrieving it with a normal get request. Our rules should deny read access to this file, but it seems to return a 200 nevertheless. Can't figure out what's the issue. Maybe it's just the emulator?

Any ideas @mkue?


mkue:
I'll have a look this week👍🏼

andrashee:
I think it's indeed that the url provided in the metadata of the uploadResponse are indeed just not protected in the emulator setup.
I could reproduce the protection when creating the downloadUrl as referenced in the GitHub issues I'm mentioning in the code. I think with this piece of code, we have an option to upload a file from the functions as admin, while keeping the storage security rules to read,write = false. Only with the given token, a client sdk or http request can download the files.

mkue (approved):
Very cool👌  I quickly addressed what we discussed before.

andrashee:
thanks for fixing it 🎉

functions/tsconfig.json
```diff
@@ -15,5 +15,5 @@
 		"target": "es2017",
 		"types": ["jest", "node"]
 	},
-	"include": ["src/index.ts"]
```
mkue, line 18: @andrashee, fyi: this fixes the error message in WebStorm

functions/tsconfig.json
```diff
@@ -15,5 +15,5 @@
...
+	"include": ["src/**/*"]
```
mkue, line 18: @andrashee, fyi: this fixes the error message in WebStorm

mkue: but it breaks the tests, it seems🤯

mkue: The `scripts/` folder seems to be the issue. I moved it out of the `src/` directory now. But we should anyway either refactor that coder or remove it.

PR #171, opened 2022-11-20
Move Firebase Admin Utils to Shared

andrashee:
The website will also use the firebase admin functions. I'm moving these to the shared directly. I added the test framework also for the shared directory, together with a github action.


ssandino (approved):

ssandino (approved):

renestalder (approved):
lgtm, but I also can't say much about that part of the code.

PR #173, opened 2022-11-21
Adding lighter subset fonts

ssandino:
- to be used on the new website for faster loading


renestalder:
@ssandino Did you on purpose add HTML files and not font files and if so, what are those HTML files for?

ssandino:
@renestalder thanks for the catch (already the second time I upload by mistake the html files with the same name)

renestalder (approved):

andrashee (approved):
👍

PR #175, opened 2022-11-25
[Website] Website Base Setup

andrashee:
Basic setup for our nextjs website.

Will create a few follow-up PRs on top of this to add more setup parts:
- [x] jest framework for unit tests
- [x] localization framework
- [x] firestore integration
- [x] deployment part


website/tsconfig.json
```diff
@@ -0,0 +1,22 @@
+{
+	"compilerOptions": {
+		"target": "es5",
```
renestalder, line 3: @andrashee Would probably go to at least es6 or better something like es2018 or es2020.

renestalder:
(Note, the review request of GitHub is sent as soon as you assign reviewers, no matter if it is in draft state. I just added a comment and then saw that this is only a draft. Sorry for that.)

PR #176, opened 2022-11-25
[Website] Adding jest and example tests: 2/n

andrashee:
As described in https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler

- jest framework for unit tests. Added an example unit test for the test middleware
- snapshot regression unit test. https://jestjs.io/docs/snapshot-testing. Compares auto generated baseline html snapshot to the current version. One can update the snapshots by a) deleting the snapshot file or b) running `npm run test:update`.

In a later step for proper e2e testing, we could add something like playwright?


renestalder:
I wouldn't necessarily put an `npm install` in front of every `make` command and rather add a `make setup` or `make install` command or something instead, so the individual workflow commands are as fast as possible, but that's more a convenience detail and how I use it.

Looks good to me, anyway.

renestalder (approved):

andrashee:
> I wouldn't necessarily put an `npm install` in front of every `make` command and rather add a `make setup` or `make install` command or something instead, so the individual workflow commands are as fast as possible, but that's more a convenience detail and how I use it.
> 
> 
> 
> Looks good to me, anyway.

I absolutely agree. Will change it in a follow-up pr for all services.

mkue (approved):
nice👍🏼

PR #182, opened 2022-12-04
refactor(general): update firestore.rules for payments access

mkue:
I updated the `firestore.rules` file so that the recipient app has proper access to the payments subcollection.

@ssandino, with this change, organisation administrators have access to the payments of their recipients. It should fix the "Payment Confirmations" view. I assume it's intended/desired that the organisation administrators are allowed to view the payments of their recipients?

Screenshot 2022-12-04 at 13 01 32


firestore.rules
```diff
@@ -67,11 +74,18 @@ service cloud.firestore {
...
+      allow read: if isGlobalAdmin(request.auth.token.email) || isGlobalAnalyst(request.auth.token.email) || canAccessRecipient(request.auth.token.email, resource.data);
       allow create: if isAdminUser(request.auth.token.email);
-      allow update: if isGlobalAdmin(request.auth.token.email) || canAccessRecipient(request.auth.token);
+      allow update: if isGlobalAdmin(request.auth.token.email) || canAccessRecipient(request.auth.token.email, resource.data);
       allow delete: if isGlobalAdmin(request.auth.token.email);
+
+      match /payments/{paymentId} {
+          allow read: if canAccessPayment(request.auth.token, recipientId);
+          allow create: if isAdminUser(request.auth.token.email);
+          allow update: if isGlobalAdmin(request.auth.token.email) || canAccessPayment(request.auth.token, recipientId);
```
mkue, line 86: @ssandino, who should be allowed to update the payments? Only global admins? Organisations admins? Or everyone who as access to the payment (including the recipient)?

MDikkii: On the app recipient should be able to change the status (e.g. set status that payment was not received). I'm not sure if it is safe to assume that app will allow only status change and we will be OK with that or should we create another document to link payment status with actual payment.

ssandino:
> I assume it's intended/desired that the organisation administrators are allowed to view the payments of their recipients?

Yes 👍

ssandino (approved):

PR #183, opened 2022-12-05
feat: iOS-Version

novas1r1:
- quick cleanup
- added missing dependencies which were only referenced
- reenabled linter
- added all missing iOS configuration


novas1r1:
@MDikkii maybe you can tell me why the pipeline is failing? Do I need to fix something? Sorry, not familiar with the whole pipeline yet :)

novas1r1:
Misses iOS icons. I need the app icon in 1024x1024px to generate it.

MDikkii:
> pipeline

    @novas1r1 Vorcel seems to be failing. It is web related only. Should be OK to not care about it here

recipients_app/lib/main_dev.dart
```diff
@@ -15,5 +15,5 @@ Future<void> main() async {
   FirebaseFirestore.instance.useFirestoreEmulator('localhost', 8080);
   FirebaseAuth.instance.useEmulator('http://localhost:9099');
 
-  runApp(MyApp());
+  runApp(const MyApp());
```
MDikkii, line 18: Are all those changes `const` related to the linter checks? Or is there other benefit of making them across whole app?

recipients_app/lib/services/authService.dart
```diff
@@ -24,15 +24,15 @@ class AuthService {
...
         verificationCompleted:
             verificationCompleted ?? (phoneAuthCredential) {},
         verificationFailed: (e) {
-          print("------- ${e.message}");
+          log("------- ${e.message}");
         },
         codeSent: codeSent ?? (verificationId, forceResendCode) {},
         codeAutoRetrievalTimeout: (e) {
-          print("autoretrievel timeout");
+          log("autoretrievel timeout");
```
MDikkii, line 35: Will it be stripped in release logs with `log` instead of `print`?

MDikkii:
@novas1r1 Vorcel seems to be failing. It is web related only. Should be OK to not care about it here

recipients_app/lib/myApp.dart
```diff
@@ -5,18 +5,22 @@ import 'package:app/view/pages/mainAppPage.dart';
 import 'package:firebase_auth/firebase_auth.dart';
 import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
+
 import 'models/currentUser.dart';
 import 'view/pages/welcomePage.dart';
 
 class MyApp extends StatelessWidget {
+  const MyApp({super.key});
```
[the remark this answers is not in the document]
novas1r1: yes. Every widget should have a key property - doesnt matter if its passed or not. Its also coming from the linter so mostly auto fixed everywhere. Here is the official rule from the dart documentation: https://dart-lang.github.io/linter/lints/use_key_in_widget_constructors.html

recipients_app/lib/main_dev.dart
novas1r1: Yes, they are auto fixed by the linter and I will give you some settings if you use VSCode so that they are auto fixed on saving the file. 

const helps to increase performance in dart. const variables (or objects) are already known in compile time and will never change. The compiler then can just reuse this object - because its never going to change - and doesnt need to create a new one.

recipients_app/lib/services/authService.dart
novas1r1: print should never be used in production since everyone could see the prints we're making. Log is just visible in debug mode. debugPrint is also fine.

https://dart-lang.github.io/linter/lints/avoid_print.html

novas1r1:
@MDikkii I also added the Icons Raphael sent me for iOS and Android :)

recipients_app/android/app/src/main/kotlin/org/socialincome/recipients_app/MainActivity.kt
```diff
@@ -0,0 +1,6 @@
+package org.socialincome.recipients_app
+
+import io.flutter.embedding.android.FlutterActivity
+
+class MainActivity: FlutterActivity() {
```
MDikkii, line 5: Why do we need second activity in `recipients_app` package?

MDikkii (commented):
ic_launcher_background / ic_launcher_foreground seems to be wrongly generated. We should have probably only vector icon for this new logo and then background color specified as background in Android Studio image icon import creator. Right now the foreground seems to have white solid background so the background file won't be visible at all.

ssandino:
> ic_launcher_background / ic_launcher_foreground seems to be wrongly generated. We should have probably only vector icon for this new logo and then background color specified as background in Android Studio image icon import creator. Right now the foreground seems to have white solid background so the background file won't be visible at all.

Added now a vector app icon without the background ([SI_Appicon_nobackground.svg](https://drive.google.com/drive/folders/1TLer0S8omiupAmryMD4GuPG-fRvTqBzf)). Hope this helps

recipients_app/android/app/src/main/kotlin/org/socialincome/recipients_app/MainActivity.kt
novas1r1: was generated automatically by flutter and should replace the java version

novas1r1:
> ic_launcher_background / ic_launcher_foreground seems to be wrongly generated. We should have probably only vector icon for this new logo and then background color specified as background in Android Studio image icon import creator. Right now the foreground seems to have white solid background so the background file won't be visible at all.

where did you check it? I did regenerate the icons but its not in testflight yet. Did you check it locally? I'm using the package in the pubspec for generating the icons for android and iOS.

recipients_app/android/app/src/main/kotlin/org/socialincome/recipients_app/MainActivity.kt
MDikkii: It seems like the app package change. We have already the MainActivity in `org/socialincome/app` package.

MDikkii: I would rather suggest to keep the one in `app` package. Do you know when exactly it was generated again?

novas1r1: when I added ios and therefore android configuration was updated again. We can remove it for now but in the future we should switch to the newer Kotlin version in general

MDikkii:
I've pulled the branch and took a look at icons, also here in the github it seems like `"foregrounds"` have white solid background. 



I haven't run the app locally as Android is throwing some errors about not supported gradle and compileSdkVersion.

recipients_app/android/app/src/main/kotlin/org/socialincome/recipients_app/MainActivity.kt
MDikkii: It is the kotlin file. Just it seems that flutter for some reason tries to use different app package names https://github.com/socialincome-san/public/tree/main/recipients_app%2Fandroid%2Fapp%2Fsrc%2Fmain%2Fkotlin%2Forg%2Fsocialincome%2Fapp

ssandino (approved):

novas1r1:
@ssandino @novas1r1 would it be fine to merge this one?

ssandino:
yes, files changed are all in the recipient_app folder, so I see lil risk that it could break something else. and no one is actively using the app. so this seems to be the safest merge since a long time ;-)

MDikkii (approved):

novas1r1:
iOS build in codemagic finally went through :)!

PR #187, opened 2022-12-09
Create CONTRIBUTING.md

ssandino:
A contributing guideline on top-level is best practice and recommended community standard for open source projects. In our case it makes sense to have one, but referring to the lower level contributing files (some have to be created first and this file to be updated with the links).


socialincome-dev (approved):
self-approval (getting structure in place before refining content)

PR #191, opened 2022-12-09
Create SECURITY.md

ssandino:
Adding a security policy file with a instruction to report a vulnerability is best practice.


socialincome-dev (approved):
self-approval

PR #201, opened 2022-12-12
Adding path to smaller font package

ssandino:
For the Admin Tool and the UI

### DIFF
--- a/admin/src/index.css
+++ b/admin/src/index.css
@@ -1,17 +1,17 @@
 @font-face {
 	font-family: 'SoSans';
 	font-weight: normal;
-	src: url('../../shared/assets/fonts/Unica77LLWeb-Regular.woff2') format('woff2');
+	src: url('../../shared/assets/fonts/Unica77LLSub-Regular.woff2') format('woff2');
 }
 
 @font-face {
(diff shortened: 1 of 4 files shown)


renestalder (approved):
@ssandino Should we maybe document somewhere what subset the `Sub` cut has?

ssandino:
@renestalder that could be helpful. Lineto actually provided us a html-file, which displays the entire character set of the cut.

Not sure how this would run directly in our setup (when copied to the shared folder). Locally it runs fine:
Screenshot 2022-12-12 at 16 36 23

Should I add the html-files in a separate PR to the font folder?

renestalder:
@ssandino I'm confused. If this is the Basic Latin range one, what does the standard `Web` font cover?

ssandino:
The standard web version covers a lot more special characters (which we never use), in example: 

Screenshot 2022-12-12 at 17 04 28

socialincome-dev (approved):

ssandino:
@mkue I think there is a approval from you as code owner necessary ... otherwise it's blocked

andrashee (approved):
🎉

PR #202, opened 2022-12-12
feat(shared): add donation certificate email html templates

mkue:
- 3 languages


ssandino (approved):

andrashee (approved):
At some point, I think it would be nice to also include the translation framework to have less duplicity. But lgtm for now 👌

PR #204, opened 2022-12-13
[Website ]Use Playwright for E2E Tests 5/n

andrashee:
- Basic Jest Snapshot testing with the `@testing-library/react` which I added before did not work well properly with the nextjs integration of `i18next`.  
- Switched to Playwright, which allows doing all kind of end-to-end testing. For now, we use automatic visual snapshot diff testing. We can add a path to the tests. Playwright creates then the baseline snapshots PNGs for several browsers and all supported languages and compares against those in following runs.
- It creates a nice html report in case of failures with visual diffing:
image


ssandino (approved):

renestalder:
@andrashee Might be worth checking at some point if we can leverage git lfs for the reference screenshot storage.

Apart from that, nice job.

PR #219, opened 2022-12-18
[Website] Declare localized alternative sites

andrashee:
One simple version of referring Google and others to the same page in a different language is to include the alternate links. According to the docs, this seems to have the same effect as defining it in a sitemap.

- https://developers.google.com/search/docs/specialty/international/localized-versions#html

This looks then e.g. like this:
image


ssandino (approved):

ssandino:
This is a nice addition!

PR #224, opened 2022-12-18
Update CODEOWNERS

ssandino:
added the dev@ to admin and shared (otherwise I get always blocked when updating readme files)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,7 +9,7 @@ firestore.rules andras@socialincome.org michael@socialincome.org
 package.json andras@socialincome.org michael@socialincome.org rene@socialincome.org
 storage.rules andras@socialincome.org michael@socialincome.org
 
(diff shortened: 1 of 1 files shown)


socialincome-dev (approved):

renestalder:
@ssandino hmm, do we all get dev@socialincome.org emails? If so, that change is quite annoying because I now get notification emails that are not necessarily meant for me.

renestalder:
Better you add exceptions for markdown/readme files in the Codeowners file

ssandino:
> @ssandino hmm, do we all get [dev@socialincome.org](mailto:dev@socialincome.org) emails? If so, that change is quite annoying because I now get notification emails that are not necessarily meant for me.

Ah true, this triggers the notifications which is forwarded. Actually only to you and Andras. I guess, to make exceptions in the codeowners file is the clean version (let me try to do this), but regardless, is there still a need to forwarded dev@ emails?

PR #226, opened 2022-12-19
[Website] Github Actions to Run & Update Snapshot Tests

andrashee:
- Adding a Github Action to update the snapshot regression tests through a PR comment.
- Enable the execution of the snapshot regression test after the unit tests.


renestalder (approved):
@andrashee Interesting approach. Very nice.

As I'm only used to GitLab CI/CD stuff and know you can do manual running jobs there to trigger this stuff, via the click of a button, I couldn't tell what the equivalent correct solution for GitHub Actions is. So, I guess that is it here.

lgtm

andrashee:
Merging this untested code for now, so that I can test it in a follow-up PR.

PR #227, opened 2022-12-19
Update CODEOWNERS

ssandino:
- adding new owner for .md files
- @renestalder do you want to get these emails too?

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -9,7 +9,8 @@ firestore.rules andras@socialincome.org michael@socialincome.org
 package.json andras@socialincome.org michael@socialincome.org rene@socialincome.org
 storage.rules andras@socialincome.org michael@socialincome.org
 
-admin/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
-functions/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
-shared/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
+admin/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
+functions/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
+shared/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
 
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -9,7 +9,8 @@ firestore.rules andras@socialincome.org michael@socialincome.org
...
 storage.rules andras@socialincome.org michael@socialincome.org
 
-admin/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
-functions/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
-shared/ andras@socialincome.org michael@socialincome.org dev@socialincome.org
+admin/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
+functions/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
+shared/ andras@socialincome.org michael@socialincome.org sandino@socialincome.org
 
+.md/ sandino@socialincome.org andras@socialincome.org michael@socialincome.org
```
renestalder, line 16: @ssandino That except a folder `.md/`.

I think you have to check the example here regarding file extensions.

renestalder:
@ssandino 

> do you want to get these emails too?

I guess I would get them anyway in those folder e.g. ui where I'm owner and not for those where I'm not owner, so that's fine.

CODEOWNERS
renestalder: [do you want to get these emails too?](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file)

ssandino:
.md should do it, right? (without the slash). 

Based on the example: 

> Order is important; the last matching pattern takes the most
> # precedence. When someone opens a pull request that only
> # modifies JS files, only @js-owner and not the global
> # owner(s) will be requested for a review.
> *.js    @js-owner #This is an inline comment.

renestalder:
Okay, let's test that and see if it works.

renestalder (approved):

ssandino:
> Okay, let's test that and see if it works.

@renestalder works (see #228)

PR #232, opened 2022-12-20
refactor(admin): remove orange money list collection and add direct csv download

mkue:
With this change we don't need the om-list collection anymore because the CSV file is created dynamically through a function.


ssandino (approved):

mkue:
I merged this to test the changes with Sandino. @andrashee, feel free to still review this and I'll address your comments in a separate PR.

PR #233, opened 2022-12-21
Update CODEOWNERS

ssandino:
Adding code owners for mobile app folder, so the reviews are better targeted (@novas1r1 @MDikkii –> you might have to link your socialincome email address to your existing GitHub account, so GitHub knows it is you. If it doesn't work, we can use also replace it with your usernames (but with the email we have it all the same way)


socialincome-dev (approved):

ssandino:
Error will be gone as soon as V + M connected their email

PR #243, opened 2023-01-03
refactor(general): update recipients access in firestore.rules

mkue:


MDikkii (approved):
It seems fine, working on an emulator 👍 

One question - can we use something different than 0 at the very start of the phone numbers in new seeds? We have some logic to remove 0 from the front in the app as users sometimes started phone with 0 as far as I understand. 

We can rethink this 0-trimming logic later, but it is problematic here for now.

Issue #181, opened 2022-12-01
[Website]: Corporate page for new website


ssandino:
**Is your feature request related to a problem?**
There is no overview of the companies/organisations who have employees contribution to Social Income, nor is there a directory of companies/organisations who contribute themselves. 

**Describe the solution/feature**
Adding a page that lists all companies/organisations, who are the `organisations-contributors` collection. There are different reasons for adding companies, namely

- have employees contributing (we know this in the future, as they can state it during the donating process) or
- donate themselves (e.g. sponsoring software like Google, Jetbrain etc.) or
- are just very known (e.g. #21) – if people/employees search for them, they will have the CTA to be the first to contribute. 

The `organisations-contributors` collection is already in the DB and Emulator with dummy data (2 entries).

@renestalder what do you recommend to use as a basic search field with live results in the dropdown in regards to this issue? What would be the basic unstyled version? Or does it make sense to directly work with a Tailwind component like [Command Palette](https://tailwindui.com/components/application-ui/navigation/command-palettes) or a simple [combobox](https://tailwindui.com/components/application-ui/forms/comboboxes#component-886412bdec54790d0c35bb5a09086c9a) (as the unstyled version in CSS would be too much work and unnecessary, if we change it soon to a Tailwind component?).

**Additional context** 
Screenshots: 
1. Inactive View
![3Artboard 1](https://user-images.githubusercontent.com/6095849/204962692-dc97211f-8133-4c1e-9aaa-18715587b005.png)

2. Active Search Field View 
![3Artboard 1 copy](https://user-images.githubusercontent.com/6095849/204962752-11d70d50-fae3-474c-9c7c-4999dce7bae8.png)

Note: the scope of this issue is on purpose small, as serves also as "good first issue" for new open source developers.

ssandino:
Any recommendations from your side @renestalder how it could make sense to proceed in this specific case? 

The usual way to develop new pages with new functionalities is:
First, check if the components already exist in our collection of components ([Storybook](https://socialincome-san.github.io/public/?path=/story/foundations-typography--page))
- if yes, use them
- if not, use basic HTML component and we switch them out later (can also be done by the same developer in a separate PR which ads the specific component to our collection)

In this case we have a bit a specific situation as there is no basic HTML component for a search field which allows for autocomplete and search results in dropdown (like [this](https://tailwindui.com/components/application-ui/navigation/command-palettes)). How should we go forward? Do you have already a Tailwind component for that kind?

[...]

PR #247, opened 2023-01-06
Added datalist element.

soumdatta81:
###  About the PR:
I have added a datalist element to the HTML using React and filled it with data from a Cloud Firestore database. However, I am not sure if the data will be retrieved because I haven’t passed the Firebase configuration object but also at the same time I don't think that will matter. In short, I am a little confused so I am doing a PR for you folks to check this and let me know if there are any issues.

refs https://github.com/socialincome-san/public/issues/181

### DIFF
--- a/website/pages/datalist.jsx
+++ b/website/pages/datalist.jsx
@@ -1,0 +1,39 @@
+// I have added a datalist element to the HTML using React and filled it with data from a Cloud Firestore database.
+import React, { useState, useEffect } from 'react';
+import ReactDOM from 'react-dom';
+import { db } from '../firebase';
+import firebase from 'firebase';
+
+const MyDatalist = () => {
+  const [items, setItems] = useState([]);
+  const options = [
+    { value: 'Option 1', label: 'Option 1' },
+    { value: 'Option 2', label: 'Option 2' },
+  ];
+
+  useEffect(() => {
+    // Set up a connection to the Cloud Firestore database
+    const database = firebase.firestore();
+
+    // Retrieve the data from the database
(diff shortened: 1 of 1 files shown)


soumdatta81:
Kindly look into this @renestalder @mkue

andrashee (commented):
We are using Next.js for the website. 

In next.js under `/website/pages` you put the individual website pages. Website specific `components` are in the `website/components` directory.

The plan is however to extract generic, reusable UI components into the `/ui/` package. See the readme and  the existing components in there.

Since with next.js we can render pages statically at build time or on the server (instead on the client) we can use the admin firestore sdk. [Here](https://github.com/socialincome-san/public/blob/main/website/pages/transparency/finances/%5Bcurrency%5D.tsx) a dummy example how to leverage next.js `getStaticProps` to query firestore.

I would start with creating a generic datalist component (which takes the arguments as parameters without any firestore dependency) in the `ui` package and then use that one in the `website` module.

renestalder:
> I would start with creating a generic datalist component (which takes the arguments as parameters without any firestore dependency) in the `ui` package and then use that one in the `website` module.

This approach, is mainly coming from Sandino and me. I recommended creating a first version of the component with a datalist just in place where it is actually used. And then we replace it later with a component from the UI library as soon I have an advanced one ready. Just to not lose time with the tooling and get a first usable prototype up and running.

renestalder:
(updated the first comment in the merge request with the related ticket, for clarity)

soumdatta81:
Added a new PR #258

ssandino:
I close this PR as it got an update in #258

PR #255, opened 2023-01-12
chore(deps): lock-file maintenance

renestalder:
Update lock-file correctly according to latest package.json


andrashee:
Thank you!

PR #257, opened 2023-01-15
[all] Refactor Firebase App / FirestoreAdmin / StorageAdmin

andrashee:
When I was working on porting the transparency page code and was writing tests for the code, I hit the problem that the firestore related tests were flaky. Jest is running the tests in parallel (which makes sense), but the individual tests were accessing the same emulated database at the same time, leading to unpredictable side effects.

With this PR, I'm proposing to change how we use the firebase app. Instead of having a magic global app object which is initiated somewhere, I would propose to have it explicitly initialized in the root of the app and then passing it through to the places where it's needed. E.g. FirestoreAdmin or StorageAdmin. Instead of having functions there, those become now stateful classes.

This makes testing IMHO easier. For each side effectful test (e.g. because it uses firestore), we can now create a new app with a different projectId. This makes the test independent of each other and can be run in parallel.


ssandino (approved):

andrashee:
@mkue if you have a minute to check if that's a valid approach to you, that would be great!

PR #258, opened 2023-01-16
Updated the previous PR 

soumdatta81:
**Added React functional component called DataList in /ui package and implemented server-side Firestore data retrieval with getStaticProps.**


ssandino:
Thanks @soumdatta81 

@renestalder everyone with our Vercel credentials can authorize the deployment (this just changed yesterday, as we had to made changes to our Vercel account). I assume this is momentarily (@andrashee knows more)

I authorised it, however the deployment failed for following reason (you can also click above in the checks to see that): 
Screenshot 2023-01-16 at 16 44 36

renestalder (changes requested):
That's a good start. There are a couple of changes that are needed:

1. Move the component from `ui/datalist.jsx` to `ui/src/components/datalist/Datalist.tsx` (use tsx and PascalCase for naming the component file).
2. Add proper TypeScript types for the component parameter, so there are easier to use. Check other components, like the [button component](https://github.com/socialincome-san/public/blob/main/ui/src/components/button/Button.tsx), for an example of how that works.
3. Prefix the component's name with `So` e.g. `SoDataList` to ensure we don't have a clash with the native HTMLDataListElement.
4. Add a Storybook Story to the same folder with the component to ensure this component is documented. See also the button component and [`button.stories.tsx`](https://github.com/socialincome-san/public/blob/main/ui/src/components/button/Button.stories.tsx) for an example of how that works.

ui/datalist.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+const DataList = ({data}) => {
```
renestalder, line 3: Add prefix and export:

```suggestion
export const SoDataList = ({data}) => {
```

ui/datalist.jsx
```diff
@@ -0,0 +1,13 @@
...
+    return (
+        <datalist id="data-from-firestore">
+            {data.map((item, i) => (
+                <option key={i} value={item.value}>{item.label}</option>
+            ))}
+        </datalist>
+    )
+}
+export default DataList;
```
renestalder, line 12: Remove `default` export. We want to embrace export barrels, and default exports make this more difficult to maintain.

```suggestion
```

renestalder (changes requested):

soumdatta81:
I am unsure of how to properly and where to connect the data list to the Firebase production collection. Can you please provide guidance or assistance in this matter?

soumdatta81:
Some changes were made with #262 and #263. I hope these will work.

PR #263, opened 2023-01-18
Added firebase config

soumdatta81:
This should do the work for the firebase configuration.


ui/src/services/firebase.ts
```diff
@@ -0,0 +1,18 @@
+//imported the Firebase module 
+import firebase from 'firebase/compat/app';
+
+
+//initialized the firebase config
+firebase.firestore().collection('firebase.json').add({
```
andrashee, line 6: Thanks @soumdatta81.

We already have this in place https://github.com/socialincome-san/public/blob/ac514a78ea00fa944f04b28f2d125b491316f254/shared/tests/FirestoreAdmin.test.ts#L1-L0

You can see here https://github.com/socialincome-san/public/blob/ac514a78ea00fa944f04b28f2d125b491316f254/website/pages/transparency/finances/%5Bcurrency%5D.tsx#L34 an example how to use it.

PR #274, opened 2023-01-21
[website] Apple Pay Check

andrashee:
Mini example how to check if the browser / device supports Apple Pay


ssandino (approved):
Great the info is available

mkue (approved):

PR #277, opened 2023-01-21
Add send sms standalone functionality

brennerthomas:


brennerthomas:
wrong base, new pull request does the same

PR #280, opened 2023-01-21
feat(recipients_app): added reusable design components in the app

novas1r1:
Added UI components based on the given AdobeXD document for:
- big buttons
- small buttons
- text styles
- success, warning and error icon
- input field and input area
- added spacing

Also updated to Flutter 3.7 because design system somehow changed there and should directly be reflected in those components.


recipients_app/lib/view/components/income/income_page.dart
```diff
@@ -25,7 +24,7 @@ class IncomePage extends StatelessWidget {
     return Consumer<CurrentUser>(
       builder: (context, currentUser, child) {
         return Padding(
-          padding: edgeInsetsAll12,
+          padding: const EdgeInsets.all(12),
```
MDikkii, line 27: Should we change it to one of the defined `app_sizes` or do we keep it like this for now as we will do another round of the redesign?

recipients_app/lib/theme/theme.dart
```diff
@@ -1,68 +0,0 @@
-import "package:animations/animations.dart";
```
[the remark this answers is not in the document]
novas1r1: haha, just renamed it to match the app_ pattern :D

recipients_app/lib/view/components/income/income_page.dart
novas1r1: another round of redesign where we will define the spacings. not top prior I would say

recipients_app/lib/theme/theme.dart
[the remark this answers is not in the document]
novas1r1: since "theme" is "reserved" by flutter :)

recipients_app/lib/view/components/income/income_page.dart
MDikkii: Sounds good 👍

MDikkii (approved):

MDikkii (approved):

PR #288, opened 2023-01-22
feature(shared): add ContributionStatsCalculator

andrashee:
Towards porting the transparency page stats into the public repo.

@mkue maybe we can also change it a bit so that we can reuse it for only 1 user (e.g. having two `build` functions) instead of all of them, and then we can use it in the donation certificate. 

Or just as an inspiration, how to do the totalContributionsChfByCurrency dynamically without hard coding the currency list.


mkue (approved):
Very nice!
Yeah, let's use this in the donation certificates, that would make a lot of sense. I'll add a TODO in my PR.

ssandino (approved):

PR #289, opened 2023-01-22
refactor(functions): use i18next translations for donation certificates

mkue:
We still use the old email template for the moment (just one now and not one per each language).

Also, I added a @react-pdf/renderer example. I think we should use it instead of pdfkit going forward. It will be super useful if we want to integrate it into user dashboard on the new website.

Also, I added a bunch of `TODO` comments that I think we should discuss at some point.


shared/src/utils/translate.ts
```diff
@@ -1,22 +1,56 @@
...
+				resourcesToBackend(
+					(language: string, namespace: string) =>
+						import(path.join(__dirname, '..', '..', 'locales', language, `${namespace}.json`))
+				)
+			)
+			.init({
+				lng: language,
+				ns: namespaces,
+				interpolation: {
+					escapeValue: false,
```
mkue, line 42: @andrashee, this was the problem with the unicode characters in the PDF. Not sure if we need it in the `templates.ts` as well. I didn't add the option there.

functions/src/etl/StripeWebhook.ts
```diff
@@ -153,7 +154,7 @@ export class StripeWebhook {
 		const balanceTransaction = charge.balance_transaction as Stripe.BalanceTransaction;
 		return {
 			source: ContributionSourceKey.STRIPE,
-			created: new Date(charge.created * 1000),
+			created: new Timestamp(charge.created, 0),
```
mkue, line 157: @andrashee: We need to make sure this works when we merge this. I would hope so.

shared/src/types/admin/Contribution.ts
```diff
@@ -20,7 +20,7 @@ export enum StatusKey {
 
 export type Contribution = {
 	source: ContributionSourceKey;
-	created: Date | Timestamp; // This is a hack. Firestore returns a Timestamp, but Firecms seems to expect a Date for a date picker. Will look into this.
```
mkue, line 23: As far as I can tell, the proper typing would be `Timestamp` here. I added a `@ts-ignore` comment to the line in the admin project that complains about this not being a `Date`.

functions/src/donation_certificates/DonationCertificatePDFTemplate.tsx
```diff
@@ -0,0 +1,53 @@
+// @ts-nocheck
```
mkue, line 1: This is just an example. We don't use it. I can also remove it again. It's a good starting point though if we want to migrate.

ssandino (approved):

functions/src/donation_certificates/DonationCertificateHandler.ts
```diff
@@ -77,11 +96,152 @@ export class DonationCertificateHandler {
...
+		const header = translator.t('header');
+		const location = translator.t('location', { context: { date: new Date() } });
+		const country = translator.t(user.address?.country as string, { namespace: 'countries' });
+		const title = translator.t('title', { context: { year } });
+		const text1 = translator.t('text-1', {
+			context: {
+				firstname: user.personal?.name,
+				lastname: user.personal?.lastname,
+				city: user.address?.city,
+				year: 2022,
```
andrashee, line 144: Should this be the variable `year`?

shared/locales/README.md
```diff
@@ -0,0 +1,4 @@
+### Good to know
+
+- The countries.json files were taken from
+  https://github.com/michaelwittig/node-i18n-iso-countries
```
andrashee, line 4: 👍

shared/locales/de/donation-certificate.json
```diff
@@ -1,25 +1,11 @@
...
-	"no-contributions": "Keine Spenden",
-	"information-1": "Bitte beachten Sie für einen möglichen Steuerabzug die kantonalen Regelungen in Ihrem Steuerkanton.",
-	"information-2": "Vielen Dank für Ihre Unterstützung.",
-	"information-3": "Freundliche Grüsse",
-	"signature-1": "Geschäftsführer",
-	"signature-2": "Leiterin Finanzen",
+	"contribution": "{{ amount, currency }}",
+	"email-subject": "Social Income Spendenbescheinigung",
+	"email-body-greeting": "Hallo {{ firstname }}",
+	"email-body-text-1": "Ich sende dir anbei die Spendenbescheinigung für das Jahr {{year}}. Herzlichen Dank für dein Vertrauen und Engagement für weniger Armut und mehr Gerechtigkeit.",
```
andrashee, line 5: @ssandino Are we `per du` with our contributors?

shared/locales/de/donation-certificate.json
```diff
@@ -1,25 +1,11 @@
...
+	"email-body-text-2": "Social Income ist in der Schweiz als gemeinnützige Organisation anerkannt und steuerbefreit.",
+	"email-body-text-3": "Wenn du Fragen hast, dann melde dich einfach bei mir.",
```
andrashee, line 7: @ssandino: Shall we maybe replace`melde dich einfach bei mir.` with something like `zögere nicht dich bei mir zu melden?`? `einfach` feels a bit mundart.

shared/src/types/admin/Contribution.ts
andrashee: In the past, the problem was that the firecms admin required the Date as input, but firestore actually returns a Timestamp. Are you sure everything still works on the admin side? Would be nice to have it only as Timestamp 👍

shared/src/types/admin/OperationalExpense.ts
```diff
@@ -4,5 +4,5 @@ export type OperationalExpense = {
 	name: string;
 	type: string;
 	created: Date;
-	amount_chf: number;
+	amount_chf: number; // TODO: change to amount and add currency field?
```
andrashee, line 7: For the stats, we want to have everything in CHF. Otherwise, you need to know the exchange rate at the time of the expense. In the end, we operate in CHF.

If we want to have additional information, I would suggest adding, an amount and currency optionally to the `amount_chf`. That's also how stripe does it.

shared/src/types/admin/User.ts
```diff
@@ -10,30 +10,30 @@ export enum UserStatusKey {
 
 export type UserAddress = {
 	city?: string;
-	country?: string;
+	country?: string; // TODO: proper typing
```
andrashee, line 13: 👍 or at least define in a comment that it needs to be two-letter iso code in uppercase.

shared/src/types/admin/User.ts
```diff
@@ -10,30 +10,30 @@ export enum UserStatusKey {
...
 		phone?: string;
-		referral?: string;
+		referral?: string; // TODO: proper typing (enum?)
 	};
 	address?: UserAddress;
 	email: string;
 	status?: UserStatusKey;
 	stripe_customer_id?: string;
-	test_user?: boolean;
+	test_user?: boolean; // TODO: discuss if still needed
```
andrashee, line 32: Is needed (and I think is useful) to filter out test user in stats.

shared/src/types/admin/User.ts
```diff
@@ -10,30 +10,30 @@ export enum UserStatusKey {
...
 	institution?: boolean;
-	language?: string;
-	location?: string;
-	currency?: string;
+	language?: string; // TODO: proper typing
+	location?: string; // TODO: discuss if still needed
```
andrashee, line 35: To be clarified with @ssandino.
At the moment, this is the most important location field for a user. This is set when a user goes through the donation process. `personal.address.country` is less often populated. Something to clarify!

shared/src/utils/translate.ts
```diff
@@ -1,22 +1,56 @@
...
-		.init({
-			lng: language,
-			ns: namespace,
+interface TranslatorProps {
+	language: string;
+	namespaces: string[];
+}
+
+export class Translator {
+	language: string;
```
andrashee, line 19: nit: maybe readonly?

shared/src/utils/translate.ts
andrashee: Ouu nice catch. 👍 
Let's add it maybe also there? Then we have a consistent behaviour?

shared/tests/utils/translate.test.ts
```diff
@@ -0,0 +1,34 @@
+import { Translator } from '../../src/utils/translate';
+
+describe('Test translations', () => {
```
andrashee, line 3: nice 👍

andrashee (approved):
Awesome stuff @mkue! Thanks a lot, looks very good now.

Just some minor comments.

functions/src/donation_certificates/DonationCertificateHandler.ts
```diff
@@ -27,38 +31,53 @@ export class DonationCertificateHandler {
...
 						if (sendEmails) {
-							const emailTemplate = getEmailTemplate(user.language);
 							await sendEmail({
 								to: user.email,
-								subject: locales['email-subject'],
-								content: handlebars.compile(fs.readFileSync(emailTemplate, 'utf-8'))({
-									firstname: user.personal?.name,
-									year,
+								subject: translator.t('email-subject'),
+								// TODO: Use renderEmailTemplate() instead of renderTemplate()
```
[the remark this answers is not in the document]
mkue: Yes, exactly!

functions/src/donation_certificates/DonationCertificateHandler.ts
mkue: Oh my, yes, of course! Good catch! :)

shared/src/types/admin/User.ts
mkue: Yes, definitely. I wonder if we really need both...

mkue:
Oops, I mistakenly enabled auto-merge yesterday🙈 @andrashee, thanks for reviewing! I address your comments in a new PR👍🏼

shared/src/types/admin/User.ts
ssandino: The address (incl. `country field`) is currently ONLY used for the donation certificate. I think it makes sense to keep a `country field` as part of the address entity. If we delete it, we run into new issues. For example: 

- Contributor has a foreign credit card, but lives currently Switzerland (i.e. a few from the staff). Kind of the expat situation. They might prefer using DE, UK, ZA as a main location (possible reasons: preferring to see EUR instead of CHF, or the future comparisons like home country to SL). But they still want a donation certificate which is sent to a Swiss address.
- Contributor from another country wants a certificate in the end of the year (we have this from Austria and Germany), as the tax authority "may" just accept a donation to a Swiss NGO or they need it for a matching with their company. 

That's why: we can't really say (a) the `country field` in an address is always _Switzerland_, (b) nor can we derive it from the `main location`. Therefore I would say: Let's keep it.

ssandino:
There were too other comments which I can't find anymore, but in short: 
**Text change:** "einfach" too informal. OK, I can find a new formulation when doing the tests and sending out the certificate to the staff. 
**Duzen:** Yes, we try to use only "du" when addressing a contributor (website, emails). Exception: donation certificate because the final recipient is the tax authority. The question will come up again with the next language addition, which is most likely French (as they seem to be less used to the informal), let's see then.

PR #295, opened 2023-01-25
feat(recipients_app): Add crashlytics for ios and android

MDikkii:
Added crashyltics for both android and iOS. 

Seems to work:
https://console.firebase.google.com/project/social-income-staging/crashlytics/app/ios:org.socialincome.app.stage/issues?state=open&time=last-seven-days&tag=all&sort=eventCount
https://console.firebase.google.com/project/social-income-staging/crashlytics/app/android:org.socialincome.app.stage/issues?state=open&time=last-seven-days&tag=all&sort=eventCount

Not sure about iOS dSym files though. As we have custom flavors it seems that default script doesn't work for us. After some changes it seems that it should upload it successfully with `build-phase` option to make it async during each build. 

Maybe we should do it only for release builds? I don't fully understand why there are missing dSym and what is the difference between optional and required 🤷‍♂️


ssandino (approved):

novas1r1 (approved):

novas1r1:
> Added crashyltics for both android and iOS.
> 
> Seems to work: https://console.firebase.google.com/project/social-income-staging/crashlytics/app/ios:org.socialincome.app.stage/issues?state=open&time=last-seven-days&tag=all&sort=eventCount https://console.firebase.google.com/project/social-income-staging/crashlytics/app/android:org.socialincome.app.stage/issues?state=open&time=last-seven-days&tag=all&sort=eventCount
> 
> Not sure about iOS dSym files though. As we have custom flavors it seems that default script doesn't work for us. After some changes it seems that it should upload it successfully with `build-phase` option to make it async during each build.
> 
> Maybe we should do it only for release builds? I don't fully understand why there are missing dSym and what is the difference between optional and required 🤷‍♂️

Should be fine for release builds only. I honestly don't know whats behind this, I uploaded them ones and then everything worked well and never cared about it anymore :D

PR #297, opened 2023-01-28
feature[admin] add playwright test

andrashee:
Our admin tool is getting a crucial part of our workflows.
To detect regressions early, I would like to add some basic testing with playwright.
The current tests cover
- authentication part in the emulator
- loading of index page
- loading of contributors list, editing a contributor and adding a new contribution
- loading of recipients list, filtering, editing a recipient and adding a new payout


renestalder (commented):
I recommend adding `data-testid` or similar attributes to the elements you are going to use for the tests (specific buttons and form fields) and select these in the tests scripts, to ensure the tests don't break for syntax reasons.

Example: `data-testid="firstname-field"`

andrashee:
> I recommend adding `data-testid` or similar attributes to the elements you are going to use for the tests (specific buttons and form fields) and select these in the tests scripts, to ensure the tests don't break for syntax reasons.
> 
> Example: `data-testid="firstname-field"`

Very good point. I need to see how we can add those in the firecms admin, but I assume that one can add additional attributes. Ok for you to merge it till then and reiterate over it?

renestalder (approved):

mkue (approved):

PR #302, opened 2023-01-30
Basic website setup with all pages 

ssandino:
Rebuilding the website structure of the current website in the new setup. I played around with the _common_ translations and added some translation that will be reused often (buttons, sections etc.).

Please note: The list of all pages on the `index page` is temporary, but helps to quickly navigate while testing (until the navigation and footer is included).


ssandino:
@andrashee I mentioned the "common" in the `useTranslation` and `getStaticProps` but didn't work yet. Let's discuss later (seems to be a small thing).

andrashee:
Very nice @ssandino. The array solution didn't work for me as well, pushed a workaround how we could use multiple namespaces.

ssandino:
> Very nice @ssandino. The array solution didn't work for me as well, pushed a workaround how we could use multiple namespaces.

Thanks! As for now it's good to have a working example with multiple namespaces (header and footer might have also their own, let's see). I marked it ready for review.

andrashee (commented):
Very nice, thanks a lot for going through all this!

I changed again how we can use multiple namespaces and updated the readme.

There is still the test failing (you can click on the red cross in the last commit and click on the failing test and go to details).

image

image

What is failing is the check if the translation files are in sync with the website.
Here the link to the translation part in the readme https://github.com/socialincome-san/public/blob/26b714c85f20a90e429f509f9fc27fe2d7be3a70/website/README.md#translations. 
What you can do is run `npm run website:extract-translations`. This will go through the code and will update the jsons. As a process, what you can do is to add the translation in the code (adding the `t(xyz)`) and then run this command to get them into all the jsons.

What do you think of adding the `common` into `website-common`?

ssandino:
Thx for the detailed information. The command `npm run website:extract-translations` works fine – it's good to clean up. As it overwrites all existing translations it's not a command for everyday use. More like a for a clean reset. Now I did it once to try out and re-added all translations. (The last version didn't actually pull the translation onto the website, so we still have the common.xy variables showing). 

Yes for renaming common to website-common. Did try, but I still get an error in the createConfig that the common.json doesn't exist. I guess it asks for it deeper in the system. So I didn't commit these changes.

ssandino:
I returned PR to _draft_ and keep playing around with it today. (that's how I learn ;-)

ssandino:
`npm run website:extract-translations` is now working smoothly (after I restarted the session it didn't overwrite files anymore (I guess it was a caching issue together with the missing 'common' file). Now I kept the common file (empty) and made a new `website-common`. So all good, it now the translations work fine with the `website-common`and the `website-index ` on the same page. 

Cleaned up some things and added more subpages in the last commit.

ssandino:
Fine for me to merge – as it is anyway temporary (the current translations in `website-index` or `website-common` might be moved into a component specific file` website-navigation` or similar). But the current version serves as an example how translations are be served and the index file has a kind of temporary directory to navigate while developing. When this is merged, I could take on the task to add the actual text translation for each page, so already would have all content. (some tests still failing)

andrashee (approved):
👍

PR #305, opened 2023-01-31
bugfix(functions): fix dependencies

andrashee:
Current deployment seems to fail due to missing dependencies. Somehow, the shared dependencies are not properly added in the functions. @mkue one more reason to get rid of the workspace setup since it's not properly supported by firebase.

image


ssandino (approved):

andrashee:
That fixed the deployment. 
image

PR #314, opened 2023-02-07
feature(admin): update recipients view

mkue:


admin/.gitignore
```diff
@@ -6,3 +6,5 @@ social-income-prod-firebase-adminsdk.json
 /test-results/
 /playwright-report/
 /playwright/.cache/
+tests/playwright/sessionStorage.json
+tests/playwright/**/*darwin.png
```
andrashee, line 10: we add this so that we don't commit the screenshots generated on the mac? 👍

admin/playwright.config.ts
```diff
@@ -23,6 +23,9 @@ const config: PlaywrightTestConfig = {
 	webServer: {
 		command: 'vite serve',
 		port: 3000,
+		env: {
+			VITE_PLAYWRIGHT_TESTS: 'true',
```
andrashee, line 27: nicely done 👍

admin/src/App.tsx
```diff
@@ -8,29 +8,33 @@ import {
...
 import { getApp } from 'firebase/app';
 import { connectFunctionsEmulator, getFunctions } from 'firebase/functions';
+import { RecipientsView } from './views/RecipientsView';
 import { ScriptsView } from './views/Scripts';
 
 const onFirebaseInit = () => {
+	const auth = getAuth();
+	if (import.meta.env.VITE_PLAYWRIGHT_TESTS === 'true') {
+		console.log('Running Playwright tests, using session storage for persistence');
+		auth.setPersistence(browserSessionPersistence);
```
andrashee, line 34: 👍

admin/src/collections/RecipientsRecentPayments.tsx
```diff
@@ -0,0 +1,84 @@
...
+	StringPropertyPreview,
+} from '@camberi/firecms';
+import { Recipient, RECIPIENT_FIRESTORE_PATH } from '@socialincome/shared/src/types';
+import { getMonthIDs } from '@socialincome/shared/src/utils/date';
+import { CreateOrangeMoneyCSVAction } from '../actions/CreateOrangeMoneyCSVAction';
+import { BuildCollectionProps } from './index';
+import { paymentsCollection, paymentStatusMap } from './Payments';
+import { firstNameProperty, lastNameProperty, orangeMoneyUIDProperty, programStatusProperty } from './Recipients';
+
+let currentDate = new Date();
```
andrashee, line 16: While we are on it: shall we change it to `const`s?

admin/src/views/RecipientsView.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { Box } from '@mui/material';
+import { RECIPIENT_FIRESTORE_PATH } from '@socialincome/shared/src/types';
+import { useState } from 'react';
+import { buildRecipientsCollection, buildRecipientsRecentPaymentsCollection } from '../collections';
+
+type RecipientViewOptions = 'all' | 'recentPayments';
+
+export function RecipientsView() {
+	const authController = useAuthController();
+	const isGlobalAdmin = !!authController.extra?.isGlobalAdmin;
```
andrashee, line 13: what is the `!!` doing?

admin/tests/playwright/admin.spec.ts
```diff
@@ -1,14 +1,41 @@
...
+			for (const [key, value] of Object.entries(storage)) {
+				window.sessionStorage.setItem(key, value as string);
+			}
+		}, sessionStorage);
+		await use(context);
+	},
+});
+
+test.beforeAll(async ({ browser }) => {
+	console.log('Logging in and writing sessionStorage.json file.');
```
andrashee, line 20: @mkue: After your change `auth.setPersistence(browserSessionPersistence);`.

Would `await page.context().storageState({ path: 'storageState.json' });` also work to save it? https://playwright.dev/docs/auth#reuse-signed-in-state

And something like
```
test.describe(() => {
  test.use({ storageState: 'userStorageState.json' });

  test('user test', async ({ page }) => {
    // page is signed in as a user.
  });
});
```
to use it?

firestore.rules
```diff
@@ -7,12 +7,12 @@ service cloud.firestore {
 
     function isGlobalAdmin(email) {
       let adminUser = get(/databases/$(database)/documents/admins/$(email)).data;
-      return adminUser.is_global_admin;
+      return 'is_global_admin' in adminUser && adminUser.is_global_admin;
```
andrashee, line 10: what does this change do?

andrashee (commented):
Amazing @mkue! Loving, that the playwright tests can now reuse the auth state.

How shall we proceed? To wrap-up https://github.com/orgs/socialincome-san/projects/14/views/1?pane=issue&itemId=19340914 what is missing?

Is the `Cash Transfers` view the same as `Recent Payments` or do we still need to create that one?

admin/.gitignore
mkue: yeah, exactly. those file never need to be comitted.

admin/src/collections/RecipientsRecentPayments.tsx
mkue: yes, sure👍🏼

admin/src/views/RecipientsView.tsx
mkue: It's simply a double negation. Makes sure that the value is actually a boolean and not `undefined` if no value is passed.

admin/tests/playwright/admin.spec.ts
mkue: Unfortunately not. Playwright uses session storage (https://playwright.dev/docs/auth#session-storage) and not regular cookies (where we could use storage state).

firestore.rules
mkue: Makes sure the function doesn't crash if there is no key called `is_global_admin`.

mkue:
Yes, we still need to do the `Cash Transfers` view. But it's super straightforward now, I'll quickly do this tomorrow.

ssandino (approved):

admin/tests/playwright/admin.spec.ts
andrashee: Ahhh I see 👌

andrashee (approved):
🎉🎉🎉

ssandino (approved):

PR #321, opened 2023-02-18
Admin: deactivate playwright tests due to being flaky

andrashee:
They are currently flaky. E.g. https://socialincome-san.github.io/public/admin-playwright-report/pr-282/

I think we need to reiterate over them before re-enabling them.


MDikkii (approved):

andrashee:
I guess our prettier github action doesn't work anymore: https://github.com/creyD/prettier_action/issues/113. Having a look.

Issue #320, opened 2023-02-16
Column "Payments left" after off-boarding


ssandino:
**Describe the issue** 
The _minus 1 payment(s) left_ for former recipients is a bit confusing, as it is very similar to the _1 payment(s) left_
![Screenshot 2023-02-16 at 23 20 20](https://user-images.githubusercontent.com/6095849/219500248-974a0e5a-c3de-4407-ac7f-7712e386f1be.png)

**Solution** 
Just writing _none_ would straightforward (with a light grey pill)

What you think @andrashee?

andrashee:
good point, addressing it in https://github.com/socialincome-san/public/pull/322

PR #322, opened 2023-02-18
[feature] – Admin: better formatting of "payments left" column

andrashee:
closes https://github.com/socialincome-san/public/issues/320


andrashee:
fyi https://github.com/creyD/prettier_action/pull/112 should fix the prettier failure

mkue (approved):

PR #323, opened 2023-02-19
[feature] – Audit log functions for firestore collections

andrashee:
First part of a few PRs to keep track of changes to our firebase data.

These triggers are taking care of keeping a history of the data in `{collectionName}-history` table. 
Some examples
- an update to `recipients/aRecipient` will copy the previous value to `recipients-history/aRecipient/history/{currentDate}`. A `status=updated` is added. At the same time, it will update the `last_updated_at` property in `recipients/aRecipient` to now.

- a deletion `recipients/aRecipient` will also copy the previous value to `recipients-history/aRecipient/history/{currentDate}` with `status=deleted`

-   an update to `recipients/aRecipient/payments/xyz` will copy the previous value to `recipients-history/aRecipient/payments/xyz/history/{currentDate}`. A `status=updated` is added. At the same time, it will update the `last_updated_at` property in `recipients/aRecipient/payments/xyzt` to now.
- deletion works the same

Unfortunately, in the triggers, one can't access the auth context and therefore one doesn't know who triggered the change.

I will propose a workaround in other PRs for this.


functions/src/etl/FirestoreAuditor.ts
```diff
@@ -0,0 +1,86 @@
+import equal from 'deep-equal';
```
mkue, line 1: Could we use the `isEqual()` function from Lodash? I think it does the same thing and then we don't need an additional library...

functions/src/etl/FirestoreAuditor.ts
```diff
@@ -0,0 +1,86 @@
...
+	auditCollectionTrigger = functions.firestore
+		.document('{collectionId}/{docId}')
+		.onWrite(async (change: Change<DocumentSnapshot>, context: EventContext) => {
+			return this.auditFirestore(change, context);
+		});
+
+	/**
+	 * Triggers changes to documents of sub collections
+	 */
+	auditSubcollectionTrigger = functions.firestore
```
mkue, line 32: Is this needed?

Couldn't we have a single trigger like this one:

```ts
auditCollectionTrigger = functions.firestore
		.document('{collectionId}/{document=**}')
		.onWrite(async (change: Change<DocumentSnapshot>, context: EventContext) => {
			return this.auditFirestore(change, context);
		});
```

This should imo work, see: https://firebase.google.com/docs/firestore/security/rules-structure#recursive_wildcards

mkue: Then the case of an update in a nested subcollection would also be covered.

mkue (approved):
Super nice! Nicely tested👌 

If we can avoid an additional library for the equality check, I would prefer that.

functions/src/etl/FirestoreAuditor.ts
andrashee: Ah nice, that would be much better. Let me try that 👍

functions/src/etl/FirestoreAuditor.ts
andrashee: True, much better

ssandino (approved):

functions/src/etl/FirestoreAuditor.ts
andrashee: Works like a charm 👍

PR #325, opened 2023-02-20
[feature] – last_updated_at & last_updated_by columns in admin

andrashee:
https://github.com/socialincome-san/public/pull/323 takes care of systematically populating the `last_updated_at` for each firestore record after a change (and populating the history collection). 

This PR suggests 
- a) display the `last_updated_at` and `last_updated_by` fields in the detail view of a record. (fyi @mkue `hideFromCollection: true,` seems also to be an easy way to remove a property from the tabular view)
- b) automatically populate the `last_updated_by` during a `onPreSave` hook. Unfortunately, the auth context is not available in the Firestore triggers, so we can't populate it there. Seems to be a [known limitation](https://github.com/firebase/firebase-functions/issues/300#issuecomment-611814916). Setting the triggering user in the `onPreSave` doesn't limit anyone of maliciously altering that record when updating the record. One workaround to enforce this is to check in the security rules that the `last_updated_by` field matches the auth user. We can do this in a separate PR.

I extended the firecms `buildCollection` to `buildAuditedCollection`. Like this, we can use it as a drop-in replacement.


shared/src/types/admin/Recipient.ts
```diff
@@ -21,7 +21,6 @@ export type Recipient = {
 	email: string;
 	first_name: string;
 	gender: string;
-	updated_on: Date;
```
andrashee, line 24: @ssandino is it okay to delete this? (there will be the last_updated_at field on all collections now)

ssandino: Yes, this field was a simple workaround to see if something changed, but no longer necessary!

ssandino (approved):

mkue (approved):
Elegantly done, given the fact that this isn't possible in the back end (which is bad for a framework the size of Firebase)👍

Issue #300, opened 2023-01-30
Adding gender specific translations into language setup


ssandino:
**Is your feature request related to a problem? Please describe.** 
Translations need some gender specific spellings (i.e. adjectives in French, or salutations).

**Describe the solution/feature**
Using variants within our language setup (i18n, locize).

**Describe alternatives you've considered**
For the moment we use both forms marked with brackets. E.g. in donation certificate: 
	"text-1": "Par la présente, nous attestons auprès des autorités fiscales que {{ firstname }} {{ lastname }} `domicilié(e)` à {{ city }} a effectué en {{ year }} des dons à l’association Social Income pour un montant total de:",

andrashee:
Have you looked at https://www.i18next.com/translation-function/context? Could be a way to implement it

ssandino:
Definitely – the `context` and `count` keys would solve this issue and make our tools much more flexible (e.g. for the amount specific [impact messages](https://socialincome.org/ch/en/get-involved-individual), donation certificates). Let's discuss tomorrow.

ssandino:
male: domicilié
female: domiciliée
all others: domicilié(e)

PR #327, opened 2023-02-20
[feature] Adding gender specific translations for donation certificate

andrashee:
closes https://github.com/socialincome-san/public/issues/300

@ssandino could also be used in other keys (everywhere where I added `{ context: genderContext }`) if you want to reformulate more.


ssandino (approved):
Very nice!

mkue (approved):

PR #330, opened 2023-02-23
enum values for referral field

ssandino:
Aligned with the values used in the website (which contributors selects in the donation process)


andrashee (commented):
Very nice! It's good to have those things enumerated.

I think it would be even nicer when we add it to the type definition in the shared collection. Like this, we can use it throughout the code (e.g. in the new website) instead of just in the admin frontend.

Here an example how we do it for the RecipientProgramStatus:
1. Enum definition: https://github.com/socialincome-san/public/blob/4adbaf4f0b5f90b9d8fad5c681e4ecf515ea5f08/shared/src/types/admin/Recipient.ts#L6
2. We say that the collection has this type instead of a string https://github.com/socialincome-san/public/blob/4adbaf4f0b5f90b9d8fad5c681e4ecf515ea5f08/shared/src/types/admin/Recipient.ts#L47
3. In the admin we create the enum values: https://github.com/socialincome-san/public/blob/4adbaf4f0b5f90b9d8fad5c681e4ecf515ea5f08/admin/src/collections/recipients/RecipientsProperties.ts#L12

ssandino:
Thx 🙏 for the hint! Makes sense, I'll give it a try

ssandino:
Added the code in `users.tsx` and user.ts. Two things I was unsure: if the idea is to make a subfolder "users" and add there a `UsersProperties.ts` file (for the moment I did it directly in the `User.ts` file). I assumed we want to use PascalCase for this, as in the example we started with a capital letter. There is still one error, as "property" is not yet used. Any hints welcome.

mkue (approved):

PR #332, opened 2023-02-25
feat(recipients_app): Fetch transactions from firestore payments collection

MDikkii:


ssandino (approved):

recipients_app/lib/data/models/recipient.dart
```diff
@@ -95,7 +97,7 @@ class Recipient extends Equatable {
       imLinkInitial: imLinkInitial ?? this.imLinkInitial,
       imLinkRegular: imLinkRegular ?? this.imLinkRegular,
       nextSurvey: nextSurvey ?? this.nextSurvey,
-      transactions: transactions ?? this.transactions,
+      transactions: transactions ?? this.transactions
```
novas1r1, line 100: trailing comma should stay there :D

recipients_app/lib/data/models/recipient.dart
```diff
@@ -159,17 +161,18 @@ class Recipient extends Equatable {
...
-        {
-          "transactions": transactions!
-              .map(
-                (element) => element.toMap(),
-              )
-              .toList()
-        },
-      );
-    }
+    // transactions / payments should be written back to payments collection, not to the user object
```
novas1r1, line 164: if its finalized remove the comments

recipients_app/lib/data/models/social_income_transaction.dart
```diff
@@ -109,10 +109,11 @@ class SocialIncomeTransaction extends Equatable {
     );
   }
 
-  factory SocialIncomeTransaction.fromJson(String source) =>
-      SocialIncomeTransaction.fromMap(
-        jsonDecode(source) as Map<String, dynamic>,
-      );
+  // TODO: do we need this?
```
novas1r1, line 112: normally we would need this to decode the string we are getting from firebase to a map

recipients_app/lib/data/repositories/repositories.dart
```diff
@@ -1,2 +1,6 @@
 export "crash_reporting_repository.dart";
+export "transaction_repository.dart";
 export "user_repository.dart";
+
+const String recipientCollection = "/recipients";
+const String paymentCollection = "payments";
```
novas1r1, line 6: is this payments or transactions?

recipients_app/lib/data/repositories/transaction_repository.dart
```diff
@@ -0,0 +1,73 @@
+import "package:app/data/models/models.dart";
+import "package:app/data/repositories/repositories.dart";
+import "package:cloud_firestore/cloud_firestore.dart";
+
+class TransactionRepository {
```
novas1r1, line 5: Good job, looks very clean :)

novas1r1 (changes requested):

recipients_app/lib/data/repositories/repositories.dart
MDikkii: that's a tricky question, on API / Admin side those are payments (as we are paying recipients some money), 

on our side, we can name them however we want probably, but maybe we can be consistent with the admin / API naming

recipients_app/lib/data/models/recipient.dart
MDikkii: 👍  done

recipients_app/lib/data/models/recipient.dart
MDikkii: 👍  done

recipients_app/lib/data/repositories/repositories.dart
novas1r1: I see. I think it makes it easier for us in the end to be consistent. SInce its also called payments in the UI - should we rename the components?

MDikkii: To keep it consistent with the rest of the project naming I've renamed all transactions related stuff to payment [db53fb8](https://github.com/socialincome-san/public/pull/332/commits/db53fb81e3b7d21b624c6bd35be7edb69b832359)

recipients_app/lib/data/models/social_income_transaction.dart
MDikkii: So for getting object from firebase response I'm using `fromMap` directly (and passing id here as separate argument). `fromJson` was not used at all for now

novas1r1 (approved):

PR #333, opened 2023-02-25
Same as pull request standalone-sms-util #2

thomasbrenner-ch:
see https://github.com/socialincome-san/public/tree/standalone-sms-util-%232

Fixed issue with .gitignore and updated it to newest main branch


andrashee (approved):
🥳 nice!

shared/tests/utils/sms.test.ts
```diff
@@ -0,0 +1,29 @@
+import { TWILIO_SID, TWILIO_TOKEN } from '../../src/config';
+import { sendSms } from '../../src/utils/messaging/sms';
+
+describe('send simple Sms', () => {
```
andrashee, line 4: I guess this will not work in the github actions without having the proper env config there?

thomasbrenner-ch: Yeah correct, but is this a problme?

andrashee: No, I think that's fine 👍 .

I think all tests should pass when one runs them locally after cloning a repo.
I would suggest that we can run this test when env is configured with the proper credentials. Otherwise, the test is skipped.
I pushed a commit with two changes. Feel free to revert if you don't like it.
- adding Twilio dependencies, otherwise the code doesn't compile.
- skipping the test if the Twilio credentials are not properly set.

mkue (commented):
very nice! 🙌

PR #334, opened 2023-02-25
Survey: firestore collection

andrashee:
Adds a sub-collection to the recipient which keeps track of all the surveys of a recipient.
The idea is that once a recipient is selected, all the surveys are created.
In addition to the survey results (`data`) we add some metadata. `access_email` & `access_pw` is referring to the firebase auth user, which is only created for 1 survey. These credentials will be part of the sharable url and allow to read/update a survey.

Let's discuss next Tuesday what additional fields / states we need.


andrashee:
One of the next steps would be to work on a Admin view which gives a very easy overview of the surveys which currently need action. (e.g. sending them out, following-up, ...)

shared/src/types/admin/Recipient.ts
```diff
@@ -10,6 +10,16 @@ export enum RecipientProgramStatus {
 	Former = 'former',
 }
 
+// TODO change to using a standard like ISO 639-3.
+export enum RecipientMainLanguage {
+	Krio = 'krio',
```
mkue, line 15: Wouldn't it make sense to already use the respective values, i.e. `kri` for Krio, `men` for Mende, etc.?

See: https://iso639-3.sil.org/sites/iso639-3/files/downloads/iso-639-3.tab

mkue: Ah ok, I see the problem. We already have entries in Firestore... I guess we need a script that converts the current entries...

mkue (approved):
Nice! lgtm👌

shared/src/types/admin/Survey.ts
```diff
@@ -0,0 +1,24 @@
+import { RecipientMainLanguage } from './Recipient';
+
+export const SURVEY_FIRETORE_PATH = 'surveys';
+
+export enum SurveyStatus {
+	New = 'new',
+	Sent = 'sent',
+	Started = 'started',
```
mkue, line 8: nit: maybe `InProgress` instead?

shared/src/types/admin/Survey.ts
```diff
@@ -0,0 +1,24 @@
...
+	Completed = 'completed',
+	Missed = 'missed',
+}
+
+export type Survey = {
+	recipient_name: string; // used to address the recipient in the form. E.g. Hello XYZ.
```
mkue, line 14: Can't we just read this from the recipient doc?

shared/src/types/admin/Recipient.ts
ssandino: Currently all 'mainLanguageProperty' (First Language) values are set to Krio (best guess) or not set at all. As there were not yet any information in a local languages, these fields were not yet maintained or kept up to date. So they can be easily erased or all better overwritten with `kri`.

If it helps @andras here the ISO codes for each language: :

- Krio = _kri_
- Mende = _men_
- Temne = _tem_
- Limba = _lia_
- plus English _en_ and _other_

andrashee: Ah perfect! In that case, I will change it to those and change the values in the DB. 👍

shared/src/types/admin/Survey.ts
andrashee: 👍

shared/src/types/admin/Survey.ts
andrashee: Yes, valid point.
The thought behind it was that the user for this can only access the survey data (since we are sending around the credentials for those in the urls) and to limit therefore all the security rules to only this document to keep things relatively closed. Wdyt?

PR #335, opened 2023-02-25
Survey: function to populate surveys & security rules

andrashee:
Allows us to create all the surveys for existing recipients. The idea would be to have an automatic trigger (not yet implemented) to create the surveys once a recipient gets into the selected state.

For now, we can trigger the survey creation in the Admin / Scripts Panel.
After execution, it looks this. This is an example of an already competed survey in firestore:

image

To be discussed which surveys should be generated.


ssandino (approved):

functions/src/etl/SurveyManager.ts
```diff
@@ -0,0 +1,122 @@
...
+		recipientName: string,
+		language: RecipientMainLanguage,
+		status: SurveyStatus,
+		due_date_at: Date
+	) => {
+		const surveysCollection = this.firestoreAdmin.collection<Survey>(
+			[RECIPIENT_FIRESTORE_PATH, recipientId, SURVEY_FIRETORE_PATH].join('/')
+		);
+		const existingSurveys = await surveysCollection.listDocuments();
+		const surveyExists = existingSurveys.filter((s) => s.id === surveyName).length > 0;
```
mkue, line 97: Since you know the full path of the document, you can do this a little easier, imo.

See:
https://github.com/socialincome-san/public/blob/fcb42bc8af539f19f2a8171351d3fdd4a82831db/functions/src/admin/AdminPaymentTaskProcessor.ts#L63-L75

functions/src/etl/SurveyManager.ts
```diff
@@ -0,0 +1,122 @@
+import * as functions from 'firebase-functions';
```
mkue, line 1: As long as the surveys are manually created from the admin tool, I would move this file to the `admin/` directory that I created in my PR. What do you think?

mkue (approved):
Very very cool! 🙌

functions/src/etl/SurveyManager.ts
```diff
@@ -0,0 +1,122 @@
...
+import { rndBase64 } from '../../../shared/src/utils/crypto';
+
+/**
+ * Takes care of creating surveys for recipients
+ */
+export class SurveyManager {
+	readonly firestoreAdmin: FirestoreAdmin;
+	readonly authAdmin: AuthAdmin;
+	readonly surveysToCreate = [
+		{ name: '01-onboarding', startDateOffsetMonths: 0 },
```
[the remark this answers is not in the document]
andrashee: Good point! Changed it in this one https://github.com/socialincome-san/public/pull/348

functions/src/etl/SurveyManager.ts
andrashee: Thanks for the pointers! Better, will change

functions/src/etl/SurveyManager.ts
andrashee: 👍 will do

PR #336, opened 2023-02-25
Survey: Website 

andrashee:
This part takes care of actually rendering the survey form towards the recipients.

I started with porting the onboarding form from typeform.
- uses https://surveyjs.io/form-library/documentation/get-started-react as a react survey framework. One can play around with it here https://surveyjs.io/create-free-survey and then copy the json configuration.
- one has to provide 4 GET parameters in the UR: recipient, survey, email & password of the firebase auth user which has access to this survey only
- with the provided token in the url, it uses the firebase auth to log in as a user so that we can leverage the firestore security rules
- takes care of translation. Retrieves the language which should be used from the survey firestore document. For this prototype I added Krio and English.
- modular questions. One can reuse the same questions in multiple surveys (onboarding, check-in, ...) without code duplication
- saves the data into firestore after each question
- also saves the data into the localstore. After refreshing the browser you continue where you left.

TODOs

- [ ] proper styling
- [ ] adding all surveys
- [ ] revisit all questions

I can demo it next Tuesday!


ssandino:
Wowowowow! Really nice

andrashee:
Here a screenshot of the first question in Krio:
image

and here all onboarding questions in English

image

image

image

image

image

image

image

image

image

image

image

image


Which then generated the following firestore entry:
image

ssandino (approved):
That's a huge progress! Well we'll done.

ssandino:
Heads up: the question `noHelpTitle` will be deleted in all surveys in the future (no added value).

MDikkii:
Are we planning to store the questions somehow in the firebase as well? I'm trying to get the idea how can we use it in the mobile app.

andrashee:
> Are we planning to store the questions somehow in the firebase as well? I'm trying to get the idea how can we use it in the mobile app.

Hi Mikołaj. Sorry for the late answer!
Yes! With this PR, we are storing the data within the surveys subcollection of the recipient.

Issue #286, opened 2023-01-22
Payment Improvements: New sliding in panel with 4 steps


ssandino:
**Step by Step Process for Payment in the Sliding Panel:** 

Should be launched in the new tile here #284 (but if this issue is done first, then fine where it is and we move the button later)

**Step 1: Register CSV**
Make new button "Register Recipients" which downloads the Registration File as CSV
-> Kerrin sends original file/format to Michael as reference

**Step 2: Payment CSV**
Change existing Download Payment File
- sorting of table after OM UID

**Step 3: Create Payments**
Make Button "Payments done"
- makes a new payment for every user with status "active" and "next payday" with date 15th plus payment status "to be confirmed" 

**Step 4: Send Text message (Please Confirm SMS)**
This will be a separate issue (will be linked later and the two requirements deleted here)

1. Make a new column in each payment, which says "Please Confirm SMS" true/false
2. When sent change to "true"

**Small things:**  
- Add new "Failed Payment" as a status (which for the moment responsible person would just change then manually, if she sees a payment didn't go through)
- Delete "Contested Date"
- "Confirmation Date": make it auto-update (and probably change name to "Last Updated" (as it's not only about Confirmations)
- In the payments sublist (slide in) make the "Status" the first row in the list and sort it the other way around (newest payment on top)
<img width="1381" alt="Screenshot 2023-01-22 at 09 19 11" src="https://user-images.githubusercontent.com/6095849/213906739-e4fd2aea-c8de-4b4e-945d-1ca9323bca41.png">

**What is done manually for the moment:**
- changes user "Status" from "active" to "former" IF someone had 36 payments
- changes user "next pay date" to  "active"
(we start like this and if it gets too messy or laborious we add a new issue with solution)

ssandino:
Not sure if this is ready for this month payout (15th), but if so then we would have to change the payout sum SLE 500 in Step 3 above (see comment from Kerrin in Slack – raise due to inflation).

mkue:
Planned update to the "Create new payments" button.

Instead of creating new payments with status set to `paid`, we update (or create) the payments for the current month and set the status to `paid`, and additionally create new payments for the next month with status set to `created` (unless they will be off-boarded the following month).

mkue:
Also, as soon as the payment is set to `paid`, the orange phone number of the recipient should be copied to the payment document.

ssandino:
Last part seems to be (almost) ready and already has its placeholder: 

[...]

PR #337, opened 2023-02-26
feature(admin/functions): add payment process tasks

### LINKED ISSUE #313: Add another download csv button to recipients view
To make the payments in the Orange Money web interface, a CSV with all the registered recipients needs to be uploaded. To simplify this process, it should be possible to always download the up-to-date CSV from the admin tool.

The current button “DOWNLOAD ORANGE MONEY CSV” should be renamed to “DOWNLOAD PAYMENTS CSV”, and the new button should be called "DOWNLOAD RECIPIENTS CSV". Both should trigger the same function call. To create the CSV, I would extend the current function and pass a parameter that defines which type of CSV should be created. 

Similar to the payments CSV, the content should only contain recipients with state `Active` and `Designated`, see https://github.com/socialincome-san/public/blob/929f1b6a278e46af91f7a2ebc16b5bfa60649666/functions/src/etl/OrangeMoneyCSVCreator.ts#L19

#### Format of the new payments CSV (one entry per recipient)

```csv
Mobile Number*,Unique Code*,User Type*
33333333,1,subscriber
55555555,2,subscriber
44444444,3,subscriber
```

mkue:
Resolves #313 

(Steps 1-3 in #286)


ssandino (approved):

shared/src/types/admin/Payment.ts
```diff
@@ -1,11 +1,29 @@
...
+export enum PaymentStatus {
+	Created = 'created',
+	Paid = 'paid',
+	Confirmed = 'confirmed',
+	Contested = 'contested',
+	Failed = 'failed',
+	Other = 'other',
+}
+
 export type Payment = {
```
mkue, line 14: @MDikkii: I slightly updated the fields on the payment document. You might have to update this on your end as well (payment status values are a little different, some fields are removed, and a `comments` field was added).

functions/package.json
```diff
@@ -21,7 +21,6 @@
 		"@playwright/test": "^1.30.0",
 		"@types/imap-simple": "^4.2.5",
 		"@types/jest": "^29.4.0",
-		"@types/deep-equal": "^1.0.1",
```
andrashee, line 24: 👍

shared/src/types/admin/Payment.ts
andrashee: @mkue what does that mean for the existing values (e.g. payment status) in firestore? Will they be displayed correctly in the admin?

andrashee (approved):
Awesome stuff @mkue!
Looking forward to trying it out

shared/src/types/admin/Payment.ts
mkue: yes, all payments are currently set to 'confirmed', so it shouldn't be a problem at all.

mkue: ok, there are 10 entries that need to be updated. I'll do it by hand quickly

MDikkii: Don't we want to keep some timestamp for changes (confirmation / contest timestamp)? 

Contest reason probably can go to `comments`, right?

mkue: Yes, I would say so. The changes are now automatically logged to a subcollection, see #323. So the information is there, and if needed for some reason, we can also add the fields back.

PR #348, opened 2023-03-03
Survey: admin overview

andrashee:
A first stab on showing an overview of the surveys in the admin tool. Needs some offline discussion on the exact process / views, but the general idea would be to have 2 main views:
- one where one can see for each recipient all the surveys and their states
- one where one can see fore each recipient, the current survey and it state. Surveys which are marked as "missed" are ignored.

On each view, one can click to edit directly a survey (would that be also interesting for the payments view?) and also copy the url (final correct url needs to be done yet) for the survey into the clipboard.

image

image

image


ssandino (approved):

andrashee:
Merging this for now due to the seed changes, happy to reiterate if there comments @mkue.

PR #349, opened 2023-03-05
feature (admin): integrate notification sms for payments

thomasbrenner-ch:
add notifications for payments


ssandino (approved):

admin/src/collections/Messages.ts
```diff
@@ -0,0 +1,39 @@
...
+				sms: 'SMS',
+				email: 'EMAIL',
+			},
+		},
+		sent_at: {
+			dataType: 'date',
+			name: 'Message Date',
+			mode: 'date',
+		},
+		content: {
```
andrashee, line 26: Here, we then could add later a custom view which renders html, so that we can display the html emails.

admin/src/collections/Messages.ts
```diff
@@ -0,0 +1,39 @@
...
+			dataType: 'string',
+			name: 'Content',
+		},
+		to: {
+			dataType: 'string',
+			name: 'To',
+		},
+		status: {
+			dataType: 'string',
+			name: 'Status',
```
andrashee, line 36: Shall we model this as an enum?

admin/src/collections/Payments.ts
```diff
@@ -54,5 +54,10 @@ export const paymentsCollection = buildAuditedCollection<Payment>({
 			dataType: 'string',
 			name: 'Comment',
 		},
+		message: {
+			dataType: 'reference',
+			name: 'Payment Notification Reference',
+			path: MESSAGE_FIRESTORE_PATH,
```
andrashee, line 60: 👍

functions/.env
```diff
@@ -1,2 +1,7 @@
 POSTFINANCE_EMAIL_USER=test@socialincome.org
 POSTFINANCE_EMAIL_PASSWORD=password
+
+TWILIO_SID=ACac5dfe59a0af841c6b771835f331fabf
```
andrashee, line 4: Are those the prod ones 😆 ?
Let's reset them in twilio and use new ones?

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
 		return `Created ${paymentsCreated} payments`;
 	};
 
+	private sendNotifications = async (recipientDocs: QueryDocumentSnapshot<Recipient>[]) => {
+		let newNotificationsSent = 0;
+		let existingNotifications = 0;
+		const now = moment();
+
+		for (const recipientDoc of recipientDocs) {
```
andrashee, line 89: Do we need an `await` here to actually await the promises in the loop?
```
for await (const recipientDoc of recipientDocs) {
```

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+
+			const paymentDocRef = this.firestoreAdmin.doc<Payment>(
+				`${RECIPIENT_FIRESTORE_PATH}/${recipientDoc.id}/${PAYMENT_FIRESTORE_PATH}`,
+				now.format('YYYY-MM')
+			);
+
+			if ((await paymentDocRef.get()).exists) {
+				const paymentDocSnap = await paymentDocRef.get();
+				const payment: Payment = paymentDocSnap.data() as Payment;
+				if (!payment.message) {
```
andrashee, line 100: I like the design here that you keep track of the messages which you sent for the payments. Allows us to trigger this function as many times as we want without spamming people 👍

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+					const recipient: Recipient = recipientDoc.data();
+					const [messageSid, _messageStatus, messageContent] = await sendSms({
```
andrashee, line 102: nit: instead of an Array, I would return an Object in sendSms. Makes it easier to use since one doesn't need to know what the first, second and third element of the array means.

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+						messageRecipientPhone: '+' + recipient.mobile_money_phone.phone,
+						messageContext: {
+							content: 'This is a Test SMS',
```
andrashee, line 105: Change the content?

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+						},
+						smsServiceId: TWILIO_SID,
+						smsServiceSecret: TWILIO_TOKEN,
+						messageSenderPhone: TWILIO_SENDER_PHONE,
+						templateParameter: {
+							language: 'de',
```
andrashee, line 111: Dynamically, based on the recipient language?

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+						content: messageContent,
+						to: '+' + recipient.mobile_money_phone.phone,
+						status: 'SENT',
+						external_id: messageSid,
+					});
+
+					await paymentDocRef.update({
+						message: messageDocRef,
+					});
+					newNotificationsSent++;
```
andrashee, line 133: Since there is quite some IO involved (which can time out and so on): shall we wrap this entire block in a try catch and have a counter for fails?

shared/src/types/admin/Message.ts
```diff
@@ -0,0 +1,18 @@
...
+	content: string;
+	to: string;
+	status: string;
+}
+
+export interface SMS extends Message {
+	external_id: string;
+}
+
+export interface Email extends Message {
```
andrashee, line 15: nice design with the inheritance 👍

shared/src/types/admin/Payment.ts
```diff
@@ -17,12 +18,14 @@ export type Payment = {
 	payment_at: Timestamp;
 	status: PaymentStatus;
 	comments?: string;
+	message?: EntityReference[];
```
andrashee, line 21: since it's an array: `messages`?

functions/src/admin/AdminPaymentTaskProcessor.ts
```diff
@@ -77,6 +81,64 @@ export class AdminPaymentTaskProcessor {
...
+						type: 'sms',
+						sent_at: Timestamp.fromDate(now.toDate()),
+						content: messageContent,
+						to: '+' + recipient.mobile_money_phone.phone,
+						status: 'SENT',
+						external_id: messageSid,
+					});
+
+					await paymentDocRef.update({
+						message: messageDocRef,
```
andrashee, line 131: one thing which I don't understand: here we add 1 docRef to the message, but the type of message is `EntityReference[]`.

andrashee (commented):
Very very nice @thomasbrenner-ch! I like the design a lot.

Most important comment is regarding the api keys. Let's change them since they are now public?

admin/src/collections/Messages.ts
```diff
@@ -0,0 +1,39 @@
...
+	name: 'Messages',
+	group: 'Messages',
+	path: MESSAGE_FIRESTORE_PATH,
+	icon: 'SupervisorAccountTwoTone',
+	description: 'Lists all messages for one recipient or user',
+	customId: true,
+	properties: buildProperties<Message>({
+		type: {
+			dataType: 'string',
+			name: 'Type',
```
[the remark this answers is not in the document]
thomasbrenner-ch: Yes, done

admin/src/collections/Messages.ts
thomasbrenner-ch: I would argue against it. Depending on the provider and channel (SMS or EMail) different status might be relevant

functions/.env
thomasbrenner-ch: no only test 🙈 but I thought I added this file to .gitignore. --> doesn't seem the case 😭

functions/src/admin/AdminPaymentTaskProcessor.ts
thomasbrenner-ch: very cool idea, added that 👍

functions/src/admin/AdminPaymentTaskProcessor.ts
thomasbrenner-ch: as it is freetext, I think the language is irrelevant for that specific case :-)

shared/src/types/admin/Message.ts
thomasbrenner-ch: yeah, was Michi's idea 🤖

shared/src/types/admin/Payment.ts
thomasbrenner-ch: I think it is required like that to make a reference to the message, isn't it?

functions/src/admin/AdminPaymentTaskProcessor.ts
ssandino: Here a first proposition (the length can be up to [160 characters](https://www.twilio.com/docs/glossary/what-sms-character-limit)): 

EN: Dear {Calling Name, _else_ First Name}, you should receive your monthly Social Income ({SLE 500}) shortly. Thanks you for confirming the payment immediately.

KRIO: Diya {Calling Name, _else_ First Name}, yu fɔ gɛt yu Social Income ({SLE 500}) we yu kin gɛt ɛvri mɔnt shɔt tɛm. Tɛnki fɔ kɔnfɔm di pe wantɛm wantɛm.

Explanation: 

- I would start with a generic one (without mentioning the channel they normally use to confirm). This can be in a second step, we would need the data of their preferred channel first in the DB (is coming with the new architecture of the number fields). 
- {...} are variables, where as the amount SLE 500 could be taken from the `payment.ts` -> `PAYMENT_AMOUNT`

Dariusz385:
Proszę o odblokowanie

mkue (approved):

PR #372, opened 2023-03-16
Fix URLs for issues

Harshil-Jani:
The URLs given for the issue labels in README were incorrect. This commit fixed them for better beginner exploration.


ssandino (approved):
Thanks!

ssandino (approved):

PR #383, opened 2023-04-05
Update recipient phone contact data

mkue:
@andrashee, @MDikkii, @novas1r1: This is how I would change the phone preferences of the recipient. What do you think?


ssandino (approved):
Good for you @novas1r1 @MDikkii ?

PR #404, opened 2023-04-27
Import missing exchange rates

andrashee:
In order to properly compute the CHF value of all the payouts (which is required to calculate the amount what we still need to pay out and to match), we need exchange rate conversion to SLE & SLL throughout the entire project time.

Since last year, we have a daily cron job in place which saves this data into firestore, this code now populates all the data starting from 2020-03-01


mkue (approved):

andrashee:
- changed the cron job to also use the API using the API key. Main difference is that that endpoint returns SLE and not just SLL and the exchange rates are aligned with the ones from Google. I will delete all existing entries and replace them.
- env variable for the api key

PR #426, opened 2023-05-10
Retrieve contact information in profile screen from Firebase

novas1r1:
Contact information in Profile screen is now fetched from Firebase:
image

If there is no contactName or contactNumber in Firebase it will hide it and only display the name of the organization.
If there is no contactNumber the "call" button will be disabled.

[⚠️](https://emojipedia.org/warning/) IMPORTANT:
We need to adjust firebase security rules to make it possible for recipients to access the organization which is set as a reference in the recipient document.
@MDikkii or @ssandino any idea, how to do that correctly? Who could adjust that?


ssandino:
@mkue would be surely know best how to amend the security rule of FB

ssandino (approved):

recipients_app/lib/data/repositories/organization_repository.dart
```diff
@@ -0,0 +1,26 @@
...
+import "package:cloud_firestore/cloud_firestore.dart";
+
+class OrganizationRepository {
+  final FirebaseFirestore firestore;
+
+  const OrganizationRepository({
+    required this.firestore,
+  });
+
+  /// Fetches the user data by userId from firestore and maps it to a recipient object
```
MDikkii, line 11: Seems like copy paste from UserRepository, can we update the comment? (Or remove)

recipients_app/lib/data/models/dashboard_card.dart
```diff
@@ -88,9 +88,14 @@ class DashboardCard extends StatelessWidget {
       case DashboardCardType.contactNumberEqualsPaymentNumber:
         final recipient = context.read<AuthCubit>().state.recipient!;
 
+        final organization = context.read<AuthCubit>().state.organization!;
```
MDikkii, line 91: `organization!` can lead to NullPointerExceptions probably. Should we make it without `!` as `AccountPage` accepts `Organization?` ?

recipients_app/lib/view/pages/main_app_page.dart
```diff
@@ -26,10 +26,13 @@ class _MainAppPageState extends State<MainAppPage> {
             // TODO display error message
             if (recipient == null) return;
 
+            final organization = context.read<AuthCubit>().state.organization!;
```
MDikkii, line 29: Same here `organization!` can lead to NullPointerExceptions probably. Should we make it without `!` as `AccountPage` accepts `Organization?` ?

recipients_app/lib/data/models/dashboard_card.dart
MDikkii: fixed

recipients_app/lib/view/pages/main_app_page.dart
MDikkii: fixed

recipients_app/lib/data/repositories/organization_repository.dart
MDikkii: done

PR #462, opened 2023-06-20
refactor(website): update surveys

ssandino:


ssandino:
@andrashee added now all proofread Krio translation. If you want to have a last look and point out how we can best roll it out asap.

PR #491, opened 2023-08-02
feat(recipient_app): Updated login registration flow

MDikkii:


ssandino (approved):

recipients_app/lib/data/repositories/user_repository.dart
```diff
@@ -59,18 +59,21 @@ class UserRepository {
...
     required String phoneNumber,
-    required Function(String) onCodeSend,
+    required Function(String, int?) onCodeSend,
     required Function(FirebaseAuthException) onVerificationFailed,
     required Function(PhoneAuthCredential) onVerificationCompleted,
+    required int? forceResendingToken,
   }) async {
     await firebaseAuth.verifyPhoneNumber(
       phoneNumber: phoneNumber,
+      forceResendingToken: forceResendingToken,
```
MDikkii, line 69: Added this to make sure that the same phone can be provided twice even if `auto retrieval timeout` hasn't passed yet.

recipients_app/lib/ui/inputs/input_text.dart
```diff
@@ -37,6 +39,7 @@ class InputText extends StatelessWidget {
             // labelText: hintText,
             suffixIcon: suffixIcon,
             floatingLabelBehavior: FloatingLabelBehavior.never,
+            counter: const SizedBox.shrink()
```
novas1r1, line 42: What is this good for and why is it empty?

novas1r1 (approved):

recipients_app/lib/ui/inputs/input_text.dart
MDikkii: When the characters limit is added Flutter is adding the counter on the bottom of the text input automatically. Which is not what we want AFAIK. One of the solutions (which also make it not adding extra space to the bottom of input fields is like above.

MDikkii: Source https://stackoverflow.com/a/58819500

novas1r1: maybe add this as a comment, otherwise we will stumble over this again :D?

MDikkii: sure 👍

PR #495, opened 2023-08-08
bugfix(website): caching of transparency page

andrashee:
@mkue Like this, it works for me. It seems to me, that providing the `TransparencyPageProps` to the `TransparencyCharts` is causing the reevaluation of `getStats` in each request. Something in there makes the caching obsolete. Providing the parameters explicitly, the caching and reevaluation seems to work fine.


website/src/app/[lang]/[country]/(website)/transparency/[currency]/page.tsx
```diff
@@ -26,7 +39,12 @@ export default async function Page(props: TransparencyPageProps) {
 					<Stats.Stat.Item variant="value">{contributionStats.totalContributions}</Stats.Stat.Item>
 				</Stats.Stat>
 			</Stats>
-			<TransparencyCharts contributionStats={contributionStats} paymentStats={paymentStats} {...props} />
+			<TransparencyCharts
```
Comment (mkue) [inline website/src/app/[lang]/[country]/(website)/transparency/[currency]/page.tsx:42]: Did you realize that this is the absolute crucial change?

Without this change, the caching doesn't work. But with this change, it doesn't matter if you leave the `getStats()` function in a separate file, if wrap it inside a `cache()` call, or specify `export const revalidate = 3600;`. All these things don't matter in this case, all that matters for the caching to work is to not use the spread (...) operator on the page props. I'm puzzled😳

Obviously, specifying `export const revalidate = 3600;` makes sense; otherwise the stats are never updated. Nonetheless, it theoretically works without...

mkue (approved):
Very cool! Thanks a lot for looking into this. I don't really understand what the issue was before with using the spread operator (see my comment), but at least it works now :)

website/src/app/[lang]/[country]/(website)/transparency/[currency]/page.tsx
Comment (mkue) [reply website/src/app/[lang]/[country]/(website)/transparency/[currency]/page.tsx:42]: Oops, I only now saw your comment from above🙈😅 I'll merge the changes.

PR #498, opened 2023-08-14
Update the logo on the website

DarkMenacer:
Screenshot 2023-08-14 at 08 37 46


DarkMenacer:
The PR seems to fail some checks but I am not able to understand what changes I’m supposed to be making for them to succeed. Could you please let me know why the checks have failed?

website/src/components/logos/si-logo.tsx
```diff
@@ -6,84 +6,18 @@ export function SILogo({ className, ...props }: HTMLAttributes<SVGElement>) {
 		<svg
 			version="1.1"
 			xmlns="http://www.w3.org/2000/svg"
+			xmlSpace="preserve"
+			id="Layer_1"
```
mkue, line 10: I don't think we need an `id`.

mkue (approved):
Hi @DarkMenacer 

I had to make [these changes](https://github.com/socialincome-san/public/pull/499/files) so that the GH actions succeed or are skipped now.

Please remove the `id` field from the SVG (if needed, it can be passed as a prop). Everything else looks good. You can check out the preview here: https://public-9q88q0v39-social-income.vercel.app/en/ch

Thank you🙏

website/src/components/logos/si-logo.tsx
DarkMenacer: On it

ssandino (approved):

PR #521, opened 2023-08-22
chore(recipient_app): Update dev files for local emulators based on real staging project

MDikkii:
Updated dev flavor files for firebase integration. 
Current local emulators setup from our project is based on staging project and in the recipients_app we had some old demo / prod hybrid on our end.


anthonyray (commented):
I could successfully build the `dev` flavor of the `recipients_app` ! Thanks !

ssandino (approved):

Issue #510, opened 2023-08-15
[DevOp]: Support monorepo setup in functions deployment


mkue:
Firebase functions don't support monorepos properly. This is why we need to use relative imports and install all dependencies specifically in the `functions/` project.

By integrating this project, the problem could be solved: https://github.com/0x80/isolate-package

The goal would be to be able to specify 

```"@socialincome/shared": "^1.0.0"``` 

as a dependency in the `functions/package.json` file and use regular imports, e.g. 
```import { FirestoreAdmin } from '@socialincome/shared/src/firebase/admin/FirestoreAdmin'``` 
instead of 
```import { FirestoreAdmin } from '../../../../shared/src/firebase/admin/FirestoreAdmin'```

More info about the problem can be found here: https://github.com/firebase/firebase-tools/issues/653

DarkMenacer:
https://github.com/socialincome-san/public/pull/526

PR #526, opened 2023-08-25
Issue #510: Integrate isolate-package in functions/

DarkMenacer:


mkue:
Thanks a lot for looking into this @DarkMenacer. I will need to try and see if your changes work with the deployment (Firebase deployments sometimes fail unexpectedly). I will get back to you as soon as I have tested the deployment with your changes.

Issue #527, opened 2023-08-25
[Mobile App Feature]: Update flutter and dart version


novas1r1:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Criteria for Completion

- [X] Updated to Dart 3.x
- [X] Updated dependencies

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

PR #528, opened 2023-08-25
feat(recipient_app):  #527: updated dart, updated dependencies

novas1r1:


recipients_app/lib/view/pages/survey_page.dart
```diff
@@ -15,6 +15,35 @@ class SurveyPage extends StatefulWidget {
...
+          },
+          onPageStarted: (String url) {},
+          onPageFinished: (String url) {
+            setState(() {
+              isLoading = false;
+            });
+          },
+          onWebResourceError: (WebResourceError error) {},
+          onNavigationRequest: (NavigationRequest request) {
+            if (request.url.startsWith("https://www.youtube.com/")) {
```
MDikkii, line 38: why do we need youtube here?

MDikkii (approved):

recipients_app/lib/view/pages/survey_page.dart
novas1r1: whooops :D copy paste

Issue #562, opened 2023-09-04
[Website Feature]: Draws overview page


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

We currently don't show any verifiable information about the drafts (selection of recipients)

### Describe the solution/feature

As discussed with @CluEleSsUK, a basic start could be something like this: 

<img width="609" alt="Screenshot 2023-09-04 at 15 45 34" src="https://github.com/socialincome-san/public/assets/6095849/9b023a2e-7a03-4483-b36e-2bb8f17cf54c">

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

ssandino:
Input from @mkue to keep it simple for you:
For the future draws section, we can just hardcode the information for the launch of the website next week.

For now the information is:
_15.10.2023, Aurora Draw, 10/400

We can always add more "intelligence" later on, where we add future draws in the admin tool and they are shown automatically when the date is in the future. But for now it might be a overkill. What you think?

CluEleSsUK:
I think this makes a lot of sense for future draws; for past draws I think it would still be great to pull from the repo so it 'follows the trail' so to speak.

ssandino:
Yes, past draws should "follow the trail"

mkue:
> I think this makes a lot of sense for future draws; for past draws I think it would still be great to pull from the repo so it 'follows the trail' so to speak.

I agree. For past draws, we should link to the file on GitHub and maybe also display its content more nicely on the website.

PR #563, opened 2023-09-06
feature(website): recipients selection page

CluEleSsUK:
I've taken a first pass at the UI for the draw transparency page as per https://github.com/socialincome-san/public/issues/562

I still need to actually pull the data from somewhere - I suspect it's possible to pull it from the repo itself as a file, as everything will be in the repo and we're using server rendering (though the component is actually `'use client'` in order to use `Disclosure` from `@headlessui`).

Some outstanding thoughts
- do I need to bump the version of `shared` somewhere to get the translations? I can't get them locally, but I see it pulls in a static version (rather than exploiting a local module)
- do I need to do anything to pull in daisyUI? I saw it's in the `ui` module, but I couldn't access anything from it out-of-the-box in `website`. That might avoid the `'use client'` declaration, and make it easier to pull the data.
- the linter was happy with my changes, though I noted there is some consistency over using `'` and `"` in different places; is there a consensus on what to use where?


CluEleSsUK:
Also it's not that pretty because I'm a terrible designer ;p if anyone has suggestions, I'm all ears!

website/src/app/[lang]/[country]/(website)/selection/page.tsx
```diff
@@ -0,0 +1,193 @@
...
+import { Disclosure } from '@headlessui/react';
+import { ChevronDownIcon } from '@heroicons/react/20/solid';
+import { Translator } from '@socialincome/shared/src/utils/i18n';
+import { BaseContainer, Typography } from '@socialincome/ui';
+import * as React from 'react';
+import { useEffect, useState } from 'react';
+import { LoaderIcon } from 'react-hot-toast';
+
+export default function Page(props: DefaultPageProps) {
+	const [translator, setTranslator] = useState<Translator>();
```
Comment (mkue) [inline website/src/app/[lang]/[country]/(website)/selection/page.tsx:12]: There is a `useTranslator` hook for this :)

Comment (mkue) [reply website/src/app/[lang]/[country]/(website)/selection/page.tsx:12]: But generally, I would argue that it's better to load the translations inside a server component and pass them to the client component (see `login/page.tsx` as an example). By doing that, the statically rendered pages will contain the text. Otherwise, the text will only show after the client components have been hydrated.

website/src/components/navbar/navbar.tsx
```diff
@@ -42,6 +44,7 @@ export default async function Navbar({ lang, country }: DefaultParams) {
 					],
 				},
 				{ title: transparency, href: transparencyHref },
+				{ title: selection, href: selectionHref },
```
mkue, line 47: I think we should probably also show a dropdown for the transparency section. No need to do it in this PR though, I'll do it in another one.

mkue:
> * do I need to bump the version of `shared` somewhere to get the translations? I can't get them locally, but I see it pulls in a static version (rather than exploiting a local module)

In local development, the cache is oftentimes not updated if you update the translations. To see them, you need to stop the website, delete the `.next/` folder and then start the website again. It's a bit annoying, only happens with client-side translations, though.

---


> * do I need to do anything to pull in daisyUI? I saw it's in the `ui` module, but I couldn't access anything from it out-of-the-box in `website`. That might avoid the `'use client'` declaration, and make it easier to pull the data.

No, you should be able to just use the react-daisyui components by importing them directly from `@socialincome/ui`.  See the `navbar-client.tsx` file, there you can see how I used the `Menu` and `Dropdown` components.
The problem with those components is though that some of them only work in client components because the dot notation, e.g. `Chart.Body`, doesn't work in server components. I hope a library update will fix this at some point.


---


> the linter was happy with my changes, though I noted there is some consistency over using ' and " in different places; is there a consensus on what to use where?

I think single quotes are used everywhere except for TSX/JSX ("HTML" code of components) where double quotes are used. I didn't configure this specifically, it seems to be a prettier default (see `.prettierignore` for more details).

website/src/components/navbar/navbar.tsx
CluEleSsUK: it's okay, I've got it!

mkue (approved):
looks all good👌 I updated the components a little and added German translations

PR #599, opened 2023-10-31
updated draw config for today's draw

CluEleSsUK:
- set the count to 10
- updated to the latest version so we use an upcoming drand round and not the current one (as it's possible to know the current one!)


CluEleSsUK:
@ssandino I've bumped the count to 10 in line with the email you sent last week - let me know if this is wrong!

ssandino (approved):

PR #607, opened 2023-11-08
refactor(recipients-selection): move files and update scripts

CluEleSsUK:
- draws now parse the file name for the number of items to draw
- draws now use the `dchoose` tool rather than a github action


mkue (approved):
looks great👌🏼

PR #636, opened 2023-11-16
feature(website): refined navbar

ssandino:
Refined NavBar: Adjusted Logo Sizes and Optimized Whitespace & Text for Mobile/Web Views

Out of scope was the CTA button "Take Action".

**Comparison**
Left: old / Right: new

Screenshot 2023-11-16 at 19 27 19
Screenshot 2023-11-16 at 19 28 21
Screenshot 2023-11-16 at 19 29 00
Screenshot 2023-11-16 at 19 30 06

### DIFF
--- a/shared/locales/en/website-common.json
+++ b/shared/locales/en/website-common.json
@@ -7,7 +7,7 @@
 		"faq": "FAQ",
 		"finances": "Finances",
 		"how-it-works": "How It Works",
-		"my-profile": "My Profile",
+		"my-profile": "Login",
 		"our-work": "Our Work",
 		"our-work-description": "Learn what we do and stand for",
 		"contributors-description": "Learn about the people donating",
--- a/shared/locales/en/website-common.json
+++ b/shared/locales/en/website-common.json
@@ -7,7 +7,7 @@
 		"faq": "FAQ",
 		"finances": "Finances",
 		"how-it-works": "How It Works",
-		"my-profile": "Login",
+		"my-profile": "My Profile",
 		"our-work": "Our Work",
 		"our-work-description": "Learn what we do and stand for",
 		"contributors-description": "Learn about the people donating",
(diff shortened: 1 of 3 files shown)


shared/locales/en/website-common.json
```diff
@@ -7,7 +7,7 @@
 		"faq": "FAQ",
 		"finances": "Finances",
 		"how-it-works": "How It Works",
-		"my-profile": "My Profile",
```
mkue, line 10: The link will say "Login" now if the user is logged in, which is wrong in my opinion. We should make it dynamic based on the fact if the user is logged in. But as long that's not the case, I would say that "My Profile" works better.

website/src/components/navbar/navbar-client.tsx
```diff
@@ -104,27 +108,33 @@ export function NavbarClient(
 									<div key={index}>
 										{_.isEmpty(section.links) && section.href ? (
 											<Link href={section.href} key={index}>
-												<Button variant="ghost">
-													<Typography size="md" weight="medium">
+												<Button variant="ghost" className="py-6 text-gray-800 hover:text-black">
```
mkue, line 111: We shouldn't use colors like `black`, `white`, `green`, etc., but only those defined here: https://github.com/socialincome-san/public/blob/main/ui/src/globals.css

ssandino: replaced the mentioned colours with the colours defined in the theme, but got rid of hover effects for text, as colors in theme not sufficient for such effect.

shared/locales/en/website-common.json
ssandino: if user sees this the same text string when logged in, then "login" would be definitely confusing. changed it back to My Profile. We can discuss at a later stage to add different terms depending on login/logout state.

ssandino:
@mkue addressed pending issues

Issue #638, opened 2023-11-17
[Mobile App Bug]: payment fields get deleted on confirmation


mkue:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I noticed that payments that are confirmed in the app, as far as I can tell, overwrite some existing fields in the payments, notably the `amount_chf` field which is required for the calculation of the payment statistics. Also, the phone_number field is deleted and an `id` field, which is not part of the [Payment model](https://github.com/socialincome-san/public/blob/92d99af5f8e7310c1fe83c23fbe356a030ab017f/shared/src/types/payment.ts#L15) is added.

@novas1r1 or @MDikkii, can you confirm that this is an issue related to the mobile app?

Example query output from Firestore:

Document ID | amount | amount_chf | comments | currency | id | last_updated_at | last_updated_by | payment_at | phone_number | status
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
/recipients/{recipient_id}/payments/2023-11 | 700 | 27.88 |   | "SLE" |   | November 17, 2023 at 11:01:49 AM UTC+1 |   | November 15, 2023 at 1:00:00 AM UTC+1 | 23200000000 | "paid"
/recipients/{recipient_id}/payments/2023-11 | 700 |   | null | "SLE" | "2023-11" | November 17, 2023 at 11:02:23 AM UTC+1 | "{recipient_id}" | November 15, 2023 at 1:00:00 AM UTC+1 |   | "confirmed"
/recipients/{recipient_id}/payments/2023-11 | 700 | 27.88 |   | "SLE" |   | November 17, 2023 at 11:01:51 AM UTC+1 |   | November 15, 2023 at 1:00:00 AM UTC+1 | 23200000000 | "paid"

### Expected Behavior

Only the `status` field should be updated.

### Screenshots?

To take a look at the corrupted payment documents, you can run this query in the Firestore console:

![Screenshot 2023-11-17 at 12 59 38](https://github.com/socialincome-san/public/assets/3518809/8db8682a-f2fd-4b9f-8509-76b60f54114a)

PR #639, opened 2023-11-17
bug(recipients_app): fix payments updates from app to not override fields not used by the app

MDikkii:
It was observed in #638 that recipients app is overriding some important data in firestore during confirming the payments. 

We had `set` method used but `update` should work better for this case. 

Also the `id` which is used by the app internally is not part of the shared contract so we will remove it with excluding it from generated `toJson` method.


ssandino (approved):
@mkue

mkue (approved):
lgtm👍

PR #650, opened 2023-11-26
feature(website): basic page for article (poverty statistics)

ssandino:


ssandino:
@mkue can you remind me how we do the links within the translations?

ssandino:
@DarkMenacer thanks for the hint – this makes totally sense. Especially for paragraphs with various links. I'll looking forward to use these `array of fragments`.

The page "poverty statistic" is more or less a unlinked landing page for some google search ads around the key word of poverty (we are part of the google non-profit program and get to freely spend 10k ads credit every month). For now the goal was to make sure the links from the running ads work after the switch to the new setup. Fast solution was:  I used the typography element `dangerouslySetInnerHTML` which did the trick for the single link. 

Looking ahead, we intend to relocate these ad-supportive landing pages to a more user-friendly blog tool, simplifying interaction for non-coders.

PR #670, opened 2023-12-07
feature(website/admin): raisenow as payment category

ssandino:


ssandino:
**Admin Tool:**
Successfully tested the assignment of new categories in the Admin tool. The functionality operates as expected without issues.
Screenshot 2023-12-08 at 16 28 14

**Donators dashboard on website:**
Confirmed that the new category assignments are properly reflected on the website's payments dashboard.
Screenshot 2023-12-08 at 16 28 03

**Potential Impact on Transparency Calculations:**
@mkue, a point for your consideration: I'm not certain if these changes could affect transparency calculations. My assumption is that it wouldn't, since the calculations seem to be independent of source category.

mkue:
I wonder if this category makes sense though. Wouldn't make it more sense if the new source would be "RaiseNow".

The thing is that Stripe will most likely support Twint soon (according to you😅). What do we do then?

I think Twint should only be the source if money is directly twinted to us and not via an intermediary like RaiseNow.

What do you think?

ssandino:
@mkue you got a good point. it's more a fix for the twint campaign which is launched in these days (so people are not confused, when it says "cash" in their dashboard)

A good solution that takes into account your valid point and also be beneficial to the user would be:

**In the admin tool we use the actual source:**
- stripe
- raisenow (instead of stripe)
- cash
- ...

**In the user dashboard we show rather payment method** (this is a bit what users care about as they don't know stripe, nor raisenow):
- if source = `stripe`, then show "credit card"
- if source = `raisenow`, then we show "twint"
- if source = `cash`, then show "credit card"
 
It still not perfect, as the Apple Pay and Google Pay contributions are not shown as payment method, for this we would have to pull the data from Stripe (not sure what we get as [data](https://stripe.com/docs/api/payment_methods/object) from them for each payment)

mkue:
ok, this seems reasonable for the moment👌🏼 i'll update it.

mkue (approved):

PR #687, opened 2023-12-22
feature (website): add accodion to recipient selection faq page

DarkMenacer:


ssandino:
Looks already good. I see in the mobile view a question that is on two lines gets centered. We had the issue also in the faq. Any way we could Aline it to the left?

image
image

DarkMenacer:
> Looks already good. I see in the mobile view a question that is on two lines gets centered. We had the issue also in the faq. Any way we could Aline it to the left?
> 
> image image

Resolved

ssandino (approved):

Issue #682, opened 2023-12-20
[Mobile App Feature]: Overview of surveys in app


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Overview of the surveys for app users. This consists of following steps: 

- [x] make first drafts in Figma
- [x] discuss it amongst developers 
- [ ] implement it

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

PR #691, opened 2023-12-27
feature(recipients_app): Add overview list for surveys

MDikkii:
Added a list page for surveys and the overview entry point.
Made all dashboard screen content items scrollable and block scrolling when no scroll is needed.
Added also calculated survey name to survey dashboard card.

https://github.com/socialincome-san/public/assets/6149761/0f5a3ec8-c67e-4f5d-9b4b-80bf068b4dc2

https://github.com/socialincome-san/public/assets/6149761/a95457f4-7ec4-4bdb-843b-9c45454121b6

Closing #682


novas1r1:
@MDikkii @ssandino intuitively I would order the surveys the other way around. So the survey next to today on the top and the one longer ago on the bottom - wdyt?

MDikkii:
What with other upcoming surveys? Like for new user we will have offboarding on the top? Or do we somehow make both upcoming and past on the bottom and "the nearest" on top?

recipients_app/lib/view/widgets/survey/survey_list_card.dart
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  String _formatDate(
+    DateTime? dateTime,
+    AppLocalizations localizations,
+    String locale,
+  ) {
+    if (dateTime == null) return "";
+
+    return "${DateFormat("dd.MM.yyyy", locale).format(dateTime)}";
```
novas1r1, line 73: @ssandino does Krio also have this date format?

novas1r1 (approved):

ssandino (approved):

recipients_app/lib/view/widgets/survey/survey_list_card.dart
ssandino: It appears that the date format is consistent throughout the entire country, regardless of the language spoken. The short date format seems to be dd/MM/yyyy, as noted [here](https://nationsgeo.com/time/africa/sl/).

ssandino:
> What with other upcoming surveys? Like for new user we will have offboarding on the top? Or do we somehow make both upcoming and past on the bottom and "the nearest" on top?

As we display all surveys in the overview (past, due, and future), arranging them in chronological order seems most intuitive. Prioritizing the user experience, especially for new users, is crucial. For instance, if their onboarding survey (currently due) is at the bottom and hidden without scrolling, it could negatively impact their user journey. Hence, the order how it is in the video would be fine for me.

recipients_app/lib/view/widgets/survey/survey_list_card.dart
MDikkii: @novas1r1  Is there any way to get the date format defined in krio_intl? Or in `intl` lib itself? I can't find a proper example right away.

MDikkii: Ok, found it. We can use skeletons defined in https://api.flutter.dev/flutter/intl/DateFormat-class.html and then it takes values from `intl` definition files

PR #706, opened 2024-01-23
refactor(admin): update phone number validation

micharied:
Changed the number validation so that every country code is allowed, but only numbers with a min. length of 5 and a max length of 15 digits. Moreover, now only integers are allowed because a phone number can't be a decimal number.

Note: I couldn't find any tests for the validation. Should we add some tests? 

It's my first open-source contribution, so I am happy about every feedback. 

Greetings Micha :)

### DIFF
--- a/admin/src/collections/recipients/RecipientsProperties.ts
+++ b/admin/src/collections/recipients/RecipientsProperties.ts
@@ -40,7 +40,7 @@ export const communicationMobilePhoneProperty: Property = {
 		phone: {
 			name: 'Phone Number',
 			dataType: 'number',
-			validation: { min: 23200000000, max: 23299999999 },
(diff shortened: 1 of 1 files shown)


ssandino (approved):
Looks good to me. Final review by: @mkue

ssandino:
@micharied, just for your information: @mkue is currently traveling and will review this as soon as possible. In the meantime, please feel free to reach out at sandino@socialincome.org, and we can add you to our Slack channel.

mkue:
Cool, thank you for your contribution :)

This looks good for the moment. But in general, I think we should switch to using a string to store a phone number using the + prefix. The validity of the phone number could then easily be checked with a regex.

mkue (approved):
lgtm👌🏼

PR #709, opened 2024-01-25
feature(admin): adding questionnaire type to overview

micharied:
Added questionnaire type in NextSurveyView.tsx


mkue (approved):
Very nice!

lgtm👌🏼

PR #712, opened 2024-01-26
feature(admin): fundraising campaign

andrashee:
Basic data model to enable the creation and an administration of fundraising campaigns.

Tasks to be done to support that users can create their own fundraising campaigns (e.g. for birthdays, weddings, ...):

- [x] Create model.
- [x] Admin interface.
- [ ] Creation Form.
    - Similar to [GiveDirectly].(https://docs.google.com/forms/d/e/1FAIpQLSceckyCzSDUQz1xZgoo3LmEpzbYPhNhBCzzuomxXqxd5m8_ww/viewform)
       - Fundraising goal should be optional in my opinion.
    - Do we also need terms @ssandino?.
- [ ] Fundraising Page.
    - Accessible by link shared to the creator by email.
    - Payment button with fundraising id attached as [metadata](https://stripe.com/docs/api/metadata).
- [ ] Adjust stripe hook: increment `amount_collected_chf` of associated campaign
    - Having a count in the campaign and updating that one feels easier rather than going through all the payments and filter them every time someone is visiting the campaign site.
- [ ] Database trigger: send email to creator after creation.
- [ ] Database trigger: send email to creator after each donation.
- [ ] Database trigger: send email to creator after reaching goal.
- [ ] Optional: Fundraising Overview.

Out of scope for now:
- Editing of the campaign after creation by the creator.
   - Having a low entry barrier (no login) to create a campaign feels crucial to me.
- Get the names of contributors as a creator. All the donations are only exposed anonymously

### DIFF
--- a/admin/src/collections/recipients/RecipientsProperties.ts
+++ b/admin/src/collections/recipients/RecipientsProperties.ts
@@ -40,7 +40,7 @@ export const communicationMobilePhoneProperty: Property = {
 		phone: {
 			name: 'Phone Number',
 			dataType: 'number',
-			validation: { min: 23200000000, max: 23299999999 },
+			validation: { min: 10000, max: 999999999999999, integer: true },
 		},
 		has_whatsapp: {
 			name: 'WhatsApp',
@@ -64,7 +64,7 @@ export const mobileMoneyPhoneProperty: Property = {
 		phone: {
 			name: 'Orange Money Number',
 			dataType: 'number',
-			validation: { min: 23200000000, max: 23299999999 },
+			validation: { min: 10000, max: 999999999999999, integer: true },
 		},
 		has_whatsapp: {
 			name: 'WhatsApp',
--- a/admin/src/views/NextSurveysView.tsx
+++ b/admin/src/views/NextSurveysView.tsx
@@ -18,7 +18,7 @@ import {
 import { StringPropertyPreview } from 'firecms';
 import { DateTime } from 'luxon';
 import { useEffect, useState } from 'react';
-import { surveyStatusProperty } from '../collections/surveys/SurveysProperties';
+import { surveyQuestionnaireProperty, surveyStatusProperty } from '../collections/surveys/SurveysProperties';
 import { NoRowsOverlay } from '../components/Overlays';
 
 type RecipientSurveyDocs = {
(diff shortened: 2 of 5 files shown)


shared/src/types/campaign.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Timestamp } from './timestamp';
+
+export const CAMPAIGN_FIRESTORE_PATH = 'campaigns';
+
+export type Campaign = {
+	email: string;
+	title: string;
+	description: string;
+	amount_collected_chf: number; // automatically updated by incoming payments.
+	goal?: number;
```
mkue, line 11: This would be the amount in CHF, right? For clarity, I would name it `goal_chf` then.

mkue:
Cool stuff!😊
Makes sense to me so far👌🏼

shared/src/types/campaign.ts
andrashee: Hmm, I would have thought in the currency of the currency referenced below. 
So that we can display round numbers on the website. The detail page of a campaign would be a fixed currency. Wdyt?

ssandino:
@andrashee 

> Do we also need terms @ssandino?

Even though the risk of misuse is unlikely, I think it's a good idea to play it safe by adding some standard clauses in a simple and easy-to-understand format. 

I recommend including these clauses directly in the existing legal terms (https://socialincome.org/en/ch/terms-and-conditions). We wanted to unify anyway all legal texts (`terms of use`, `terms and conditions`, `privacy policy` and now newly `terms for fundraiser`) on one page .../legal. This could be the chance to do it. 

Good for you? If yes I would do this in another PR and you can just link to the (future) site .../legal.

**Proposal for the 5 Clauses for Private Fundraisers:** 

1. The person starting a private fundraiser on Social Income's website is responsible for setting up and running the fundraising event. This responsibility includes covering all costs, planning, promoting, and ensuring compliance with the law.
2. Social Income will create a special 'fundraising page' on its website. The fundraiser will receive a unique link to this page and should direct all supporters to donate to Social Income through this specific page. The fundraiser must not accept donations or any other payments on behalf of Social Income.
3. The fundraiser is permitted to use Social Income's name and logo to promote the fundraising event and raise money for Social Income. However, this does not imply that Social Income endorses the event. Social Income reserves the right to discontinue or refuse the use of its name and logo at any time.
4. The fundraiser understands that Social Income owns all data collected, including information gathered during the fundraiser or on the 'fundraising page.' Social Income may allow the fundraiser to use this data to thank donors.
5. The fundraiser is responsible for ensuring that all aspects of the fundraising event are appropriate and respectful. No part of the event should be harmful, offensive, untrue, or discriminatory. Social Income has the right to terminate the fundraising campaign at any time at their discretion.

ssandino:
Update on

> Do we also need terms @ssandino?

The terms for fundraisers will be introduced with #738

ssandino:
@andrashee the fundraiser terms are online and can be linked:

https://socialincome.org/en/ch/legal/fundraisers

Thx 🙏. @DarkMenacer for the great work

PR #713, opened 2024-01-26
feature(website): campaign metadata in stripe checkout

andrashee:
Preparational PR to have the option to append a campaignId to the stripe checkout

PR #714, opened 2024-01-26
feature(functions): update campaign amount_collected_chf in stripe hook

andrashee:
More an outline of how it could be done, rather than the final implementation. 
Needs testing if the metadata appended here https://github.com/socialincome-san/public/pull/713 actually gets appended to a charge by stripe.


shared/src/stripe/StripeEventHandler.ts
```diff
@@ -112,6 +113,28 @@ export class StripeEventHandler {
...
+		if (charge.metadata.campaignId) {
+			const campaignRef = this.firestoreAdmin
+				.collection<Campaign>(CAMPAIGN_FIRESTORE_PATH)
+				.doc(charge.metadata.campaignId);
+
+			try {
+				const campaign = await campaignRef.get();
+				const current_amount_chf = campaign.data()?.amount_collected_chf ?? 0;
+				await campaignRef.update({
+					amount_collected_chf: current_amount_chf + charge.amount,
```
mkue, line 129: Is it possible that you should use `charge.balanceTransaction.amount` here? The `amount` field could be in a different currency, right?

At least that's how we do it in the `constructContribution()` function.

andrashee: Ah very good point. Will change.

andrashee:
> Do you think it would make sense to update the `constructContribution()` function (and the Contribution model) and link the campaign name to the contribution object?

That's a very good point, otherwise we actually lose the reference after we dismiss the original charge object. 
Will change.

PR #715, opened 2024-01-26
feature(recipients_app): add pull to refresh

novas1r1:
Added ptr to the DashboardScreen and the PaymentsScreen to be able to update to latest data fast.
Needs to be tested with surveys, but PR of @MDikkii needs to be merged before being able to do that.


ssandino (approved):

recipients_app/lib/view/pages/payments_page.dart
```diff
@@ -7,26 +7,34 @@ import "package:flutter/material.dart";
...
     return Scaffold(
       appBar: AppBar(
         elevation: 0,
         title: Text(localizations.payments),
         centerTitle: true,
       ),
-      body: Padding(
-        padding: AppSpacings.h8,
-        child: Column(
+      body: RefreshIndicator(
```
MDikkii, line 33: Something seems broken with scrolling. Like the nested listview is not scrolling at all. I can only scroll on the header 🤔

recipients_app/lib/view/pages/dashboard_page.dart
```diff
@@ -75,32 +82,31 @@ class _DashboardView extends StatelessWidget {
...
-        return Padding(
-          padding: AppSpacings.h8,
-          child: Column(
+        return RefreshIndicator(
+          key: _refreshIndicatorKey,
+          onRefresh: () async {
+            context.read<PaymentsCubit>().loadPayments();
+            context.read<SurveyCubit>().getSurveys();
+          },
+          child: ListView(
```
MDikkii, line 91: Also here we have nested listview which seems to throw some errors when there are some surveys on home page

MDikkii (approved):

PR #716, opened 2024-01-26
feature(website): campaign page

andrashee:
Prototyping a detail page of a campaign then test the loop of a payment with a campaign metadata to stripe and back.


andrashee:
Merging to quickly check in staging how this metadata is appended to the charges on which we are listening to.

Issue #743, opened 2024-02-07
[Website Feature]: Implement animated logo


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

We currently use a static text logo on the top left of the website instead of the animated text logo. 

[Social Income Logo Animation.json (Lottie Export)](https://github.com/socialincome-san/public/files/14201388/Soical.Income.Logo.Animation.json)

You can preview the animation by uploading the json file (above) to [this site](https://drawer.design/preview-lottie-animation/).

### Describe the solution/feature

Adding the animated logo – this requires to use [Lottie React Player](https://github.com/LottieFiles/lottie-react). A good start is these articles: [The easiest way to handle animations In React/Nextjs](https://medium.com/@seun.thedeveloper/the-easiest-way-to-handle-animations-in-react-nextjs-5934a689a010) and [How To Use Lottie Animations in a React App](https://lottiefiles.com/blog/working-with-lottie-animations/how-to-use-lottie-in-react-app/)

The mobile view with the icon logo should not be changed. 

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

MoarCoding:
I'd like to work on this

ssandino:
Great @MoarCoding – thanks for your coding contribution!

MoarCoding:
@ssandino Hi again. The linked lottie has a lot of vertical whitespace and I also feel that the left side should always be aligned with the left side of the main content. Would you like me to make a PR for including the lottie and then you can adjust said lottie the way you see fit before merging into main?

ssandino:
> @ssandino Hi again. The linked lottie has a lot of vertical whitespace. I also feel that the left side should always be aligned with the left side of the main content which it only does at animation end. Would you like me to make a PR for including the lottie and then you can adjust said lottie the way you see fit before merging into main?

Good point. Thanks for thinking along. Yes a first PR without the perfect animation makes surely sense, hence, the code review can take place and we exchange the Lottie file later.

@RaphaelSchoen1 could you send me the original AF file? Then we could do a first version in which we cut all possible whitespace (that should already look alright) and upon do a new version, in which the animation only extends to the right.

MoarCoding:
@ssandino  PR up

RaphaelSchoen1:
[...]

PR #745, opened 2024-02-09
feature(website): animated logo

MoarCoding:
This replaces the top navbar's logo with a Lottie animation

Note: Not ready to be merged until lottie-file's whitespace has been pruned and the text only animates to the right.


ssandino:
@MoarCoding [here](https://public-git-fork-moarcoding-issue-743-imple-b2cfb8-social-income.vercel.app/en/ch) you can see the preview deployment. I see that the Lottie file itself is centred in its div and could be left aligned.

Screenshot 2024-02-09 at 14 46 12

MoarCoding:
@ssandino Exactly. I tried editing the lottie myself, but in the end I failed to find one that let me save the result as json. Figured it would be easier for you guys to just update it according to your wishes.

ssandino:
@MoarCoding we have now an updated version of the Lottie file thanks to @RaphaelSchoen1. See [comment in issue](https://github.com/socialincome-san/public/issues/743#issuecomment-1936994832). Would you be able to update before the review?

MoarCoding:
@ssandino I'm on it. Sorry for the delay, I only code during office hours.

ssandino:
Thanks @MoarCoding – the preview deployment has currently an error, but seems unrelated to the animated logo.

mkue:
Thanks for you work @MoarCoding 

I disabled the looping of the animation now by default. It looks weird if the animation keeps repeating.

The downside of the animated logo is that it doesn't work with statically created pages (it requires JS). This means, that the logo is blank when the page loads and then there's some flickering.

I think we should prevent this. Do you want to look into this @MoarCoding? Let me know if you need more info.

ssandino:
Discussed with @RaphaelSchoen1 if the loop can be disabled directly in the Lottie file (and therefore replace the current file), but it turns out this has to be done directly in the code: 
[see API documentation](https://github.com/lottie-react-native/lottie-react-native/blob/master/docs/api.md) -> Loop

A fallback option would be great!

MoarCoding:
> Thanks for you work @MoarCoding
> 
> I disabled the looping of the animation now by default. It looks weird if the animation keeps repeating.
> 
> The downside of the animated logo is that it doesn't work with statically created pages (it requires JS). This means, that the logo is blank when the page loads and then there's some flickering.
> 
> I think we should prevent this. Do you want to look into this @MoarCoding? Let me know if you need more info.

Absolutely. I'm away on vacation right now, but I'll be happy to look into this come monday.

MoarCoding:
@ssandino I've pushed a fallback. Still looks a bit weird IMO since the lottie's starting frame doesn't look like the fallback, but I figured that it's up to you to figure out what you're after here

ssandino:
> @ssandino I've pushed a fallback. Still looks a bit weird IMO since the lottie's starting frame doesn't look like the fallback, but I figured that it's up to you to figure out what you're after here

Thanks @MoarCoding I will try to replace the current fallback logo with the first frame (exported as SVG) of the animation. Theoretically it should then be seamless. Let me try that. 

While testing I found a weird bug on Safari. The logo itself is not left aligned. Any solution to this? 

Screenshot 2024-03-01 at 15 10 12

MoarCoding:
> > @ssandino I've pushed a fallback. Still looks a bit weird IMO since the lottie's starting frame doesn't look like the fallback, but I figured that it's up to you to figure out what you're after here
> 
> Thanks @MoarCoding I will try to replace the current fallback logo with the first frame (exported as SVG) of the animation. Theoretically it should then be seamless. Let me try that.
> 
> While testing I found a weird bug on Safari. The logo itself is not left aligned. Any solution to this?
> 
> Screenshot 2024-03-01 at 15 10 12

Can't confirm since I don't have access to safari. Does it work without my changes?

ssandino:
thanks for the hint. let me see if I find the bug. I get back to you.

ssandino:
> thanks for the hint. let me see if I find the bug. I get back to you.

The wrapping didn't work. @renestalder do you have a good idea?

renestalder:
@MoarCoding @ssandino 

1. Set the parent `<a>` to be an `inline-block`. You can't have an inline element containing block elements.
2. Remove `height:38px` from the parent `div` and set the target width on the `<svg>`. It's set to `width: 100%`. Of course it takes the whole width. ou can use `aspect-ratio` CSS property to then enforce constraints.

And then you get this:

Screenshot 2024-03-04 at 10 46 57

renestalder:
@MoarCoding @ssandino Regarding Lottie files: These files are relatively huge compared to what they do. We should make sure that stuff is loaded non-render blocking to not slow down the website with unnecessary animations on slow connections, using async `import()` or a `React.Lazy` on a higher level.

ssandino:
Thanks @renestalder (added you as reviewer). @MoarCoding would these pointers help to update the PR and wrap up the issue? As we start a large campaign later in the week, it would be amazing to get your branch deployed by Friday. Possible?

MoarCoding:
@ssandino Hi. Unfortunately I'll be extremely busy at work this week since I've just started a new assignment. Would it be possible for @renestalder to fix this himself?

PR #747, opened 2024-02-12
feature(general): mailchimp api integration

brennerthomas:
- Adding in shared the Event Handler --> can be used by all apps
- Adding a specific next.js api for Mailchimp inside the Website


mkue:
looks very good @brennerthomas!

i changed the api endpoints a bit to prevent a security issue so that users can only change their own mailchimp subscription.

also, i started with the integration in the frontend, but it's just dummy code yet. feel free to polish it if you want :)

the `useUpdateMailchimpSubscription()`, `useCreateMailchimpSubscription()` hooks take use of react-query, which is very handy if you want to refetch some data after an update has been made. if you have trouble understanding that code, i can explain it to you in person at some point :)

mkue (approved):
great stuff @brennerthomas 🙌🏼🙌🏼

ssandino:
very nice @brennerthomas 🙌🏼🙌🏼

PR #778, opened 2024-03-19
feature(website): newsletter popup

brennerthomas:
Adding a new component that can be added to any page very easily as

			<NewsletterPopup

				lang={params.lang}

				region={params.region}

				delayTime={3000}

			/>

at the moment the popup is rendered twice because of 'strict' mode


website/.env.sample
```diff
@@ -0,0 +1,21 @@
+# These environment variables are required so that the Firebase Admin SDK (server side) connects to the local emulator.
```
mkue, line 1: I would name this `.env.local.sample` and only have entries that are not in the `.env.development` file (the mailchimp stuff..)

PR #779, opened 2024-03-21
feature(website): faq section on campaign page

ssandino:


ssandino:
Please note: Once users have the ability to create their own campaigns, the content of the FAQ will be generalized. Currently, the FAQ specifically mentions Ismatu's campaign, which is unique in its purpose of raising funds for a specific cause. This mention will be removed. In the future, the FAQ will consist of approximately three questions that will apply universally to every campaign.

mkue (approved):

PR #789, opened 2024-03-30
feature(functions): first payout email function (initial version)

mkue:


mkue:
FYI, @ssandino: this is ready to be merged once the template is ready. Let me know all the template variables that you need, and then I'll update the code here.

PR #791, opened 2024-03-30
feature(website): sections for new landing page

DarkMenacer:


DarkMenacer:
@mkue @ssandino 
Worked on the quotes section. Most of it is complete but I didn't add the translations code because I have a doubt there (regarding the avatar image data). I'm waiting for now, I'll get these two reviewed and then continue working on the other sections.

website/src/app/[lang]/[region]/v2/(home)/(sections)/faq.tsx
```diff
@@ -1,25 +1,71 @@
...
-					</Typography>
+			<Typography size="2xl" className="mb-12 text-center">
+				{translator.t('section-6.title-1')}
+			</Typography>
+			<hr />
+			<Accordion type="single" collapsible className="mb-10 w-full">
+				{questions.map((question, index) => (
+					<AccordionItem key={index} value={`item-${index}`}>
+						<AccordionTrigger>
+							<Typography size="lg" className="text-left">
```
Comment (mkue) [inline website/src/app/[lang]/[region]/v2/(home)/(sections)/faq.tsx:39]: Is the "text-left" required? Seems to me that this should be the default and part of the AccordionTrigger.

mkue:
looks good to me🙌🏼

DarkMenacer:
@mkue @ssandino 
Added two more sections
overview.tsx and approach.tsx

Could you review these sections as well?

shared/locales/en/website-home2.json
```diff
@@ -92,17 +92,22 @@
 	"section-7": {
 		"title-1": "Social Income combines three approaches to change."
 	},
-	"section-8": {
+	"section-x": {
```
mkue, line 127: Was this intended?

mkue (approved):
Looks good! Let's merge it and do the next sessions in a new PR, ok? This makes it a little easier to review.

DarkMenacer:
Works

DarkMenacer:
@mkue Can you merge this branch, I'll quickly take up the next sections...
We plan to present the new landing page during the coming townhall on the 1st of May

ssandino (approved):
@DarkMenacer, I think you should be able to merge this yourself, as @mkue has already reviewed it and one approval should make it mergeable for you. Please let me know if this option is somehow not available to you.

DarkMenacer:
Okay, got it
I'll merge the PR then

DarkMenacer:
@ssandino 
Screenshot 2024-04-23 at 07 50 44

It says this
I am not able to merge it

ssandino (approved):

ssandino:
@DarkMenacer ah I know why, when things change in the UI folder @renestalder is asked an additional review. @renestalder could you click "approved" (no review required). Thanks

ui/src/components/accordion2.tsx
```diff
@@ -0,0 +1,56 @@
...
+			className,
+		)}
+		{...props}
+	>
+		<div className="pb-4 pt-0">{children}</div>
+	</AccordionPrimitive.Content>
+));
+AccordionContent.displayName = AccordionPrimitive.Content.displayName;
+
+export { Accordion, AccordionContent, AccordionItem, AccordionTrigger };
```
renestalder, line 56: Add the `export` directly to the corresponding functions/const declarations. There is no benefit of collecting them at the end of the file.

ui/src/components/accordion2.tsx
```diff
@@ -0,0 +1,56 @@
+'use client';
+
+import * as AccordionPrimitive from '@radix-ui/react-accordion';
```
renestalder, line 3: Consider using the native `<details>` element with additional group closing handlers instead of adding a dependency on top that is probably already obsolete.

We benefit from using what is already there in HTML.

ui/src/components/accordion2.tsx
renestalder, line 1: I don't like that this is just called `accordion2`. That is not a semantically useful name and I smell that there is probably already an accordion that could have been extended instead of adding another component with a not so descriptive name.

renestalder (approved):
@ssandino @DarkMenacer You might wanna remove me from all code owner assignments and code review tasks. I'm so out of the loop, looking at the code I feel like you are doing anything in the code style we have in that project, but I would anyway disagree with many things because that's not the way I'd consider it best practice. My reviews only lead to many comments that are not particular helpful in this code base. I've become a very picky code reviewer as I look very closely at things like accessibility and performance. And I lead teams with focus of sustainability of code bases, that remove Tailwind and unnecessary third-party React dependencies. Nothing of that is particilar useful for Social Income at the moment.

ui/src/components/accordion2.tsx
ssandino: @renestalder es, renaming to the standard name `accordion `is planned following the migration of the new design.

ui/src/components/accordion2.tsx
ssandino: @DarkMenacer thank you for considering this for a future PR.

ui/src/components/accordion2.tsx
ssandino: @DarkMenacer thank you for considering this for a future PR.

ui/src/components/accordion2.tsx
mkue: This is definitely fine as the code directly comes from https://ui.shadcn.com/docs/components/accordion and does not need to be changed apart from the tailwind styling...

PR #793, opened 2024-04-04
 feature(website): styling cookie and newsletter signup banner

ssandino:
- **Adapted to New Design:** Implemented the updated design specifications from Figma
- **Improved Banner Placement:** Refined the positioning logic for our notification banners. The update ensures that when multiple banners are displayed simultaneously, they do not overlap

**Future Improvements:** As new components are developed, consider removing the custom classes from the buttons and instead utilize the plain component directly in the future. This approach will streamline the codebase. However, the adjustments made in this update already represent an improvement. Therefore, I advocate for proceeding with deployment (if now code changes needed).

### Before: 
before3

### After: 
Note the white background is hardly visible in the GIF (but it's there)
after1

### DIFF
--- a/shared/locales/de/website-common.json
+++ b/shared/locales/de/website-common.json
@@ -33,7 +33,7 @@
 		"legal": "Rechtliches"
 	},
 	"cookie-consent-banner": {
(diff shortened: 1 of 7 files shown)


mkue:
Closing this, as banners are not shown at the same time anymore.

PR #798, opened 2024-04-18
feature(website): redirects legal pages

ssandino:
Updated middleware.ts to redirect old legal page URLs (e.g., /privacy) to their new paths under /legal/* (e.g., /legal/privacy). This adjustment maintains compatibility with external tools like the mobile app that still use the old URLs. Deleted the old pages.


ssandino:
@mkue, while working on this, I discovered an issue with the redirects. 

They fail to handle localizations; for instance, socialincome.org/twint redirects correctly, but socialincome.org/en/ch/twint does not. Since we have not frequently communicated about the language-specific versions (although we have used relative links within the site, I'll change them in the next commit), this issue seems not too urgent and could be optimized in the future. Let me know what you think, and I can create an issue for it.

Idea: To address this, we could modify our middleware to check for the end of the path rather than exact matches. This change would allow it to handle localized paths and other prefixes in our URL structure.

mkue (approved):

PR #802, opened 2024-04-22
(recipient selection) draw for organisation reachout salone

CluEleSsUK:


ssandino (approved):

ssandino:
@CluEleSsUK cleaned up the pr to have a clean commit with only the reachout salone list. have a look

CluEleSsUK:
Looks good, though GitHub is complaining about conflicts - you might have to fiddle with it in the GitHub editor. We‘ll need to remove testlist from the private repo so it doesn’t get recreated but we can do that next time

ssandino:
I reverted the commit that deletes the test lists. I'll do it in a separate PR.

PR #821, opened 2024-05-01
feature(website): townhall presentation

DarkMenacer:


DarkMenacer:
@mkue Did some polishing work
I've tried my best to add all of your changes however I must admit not all of them have been added
I'll look into those after the townhall

mkue (approved):

Issue #733, opened 2024-01-30
[Admin Tool Feature]: Forecast view for payouts (next 6 months)


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

It's currently difficult to assess how many payments are to be made in the coming months and, correspondingly, how much money will be needed. Since there are always some recipients being offboarded and onboarded, this number is constantly changing. However, it's crucial for assessing the required funds in the bank accounts in SL.

### Describe the solution/feature

We need to add a new tile/view under the 'finance' tab in the admin tool. This tile should provide an overview of all Social Income payments to recipients for the next six months, resulting in a list with a total of six rows.

**Example table with 3 columns (as of January 30th, 2024):**

February 2024 / x Recipients / USD x
March 2024 / x Recipients / USD x
May 2024 / x Recipients / USD x
June 2024 / x Recipients / USD x
July 2024 / x Recipients / USD x
August 2024 / x Recipients / USD x

(if current date is < 15th of the month then show the current month. Payments are always done on 15th.)

**Calculations**

`Recipients x` should count all `active recipients` of the respective month (calculated based on starting month) + all `starting next payday` (add them in current month if today is < 15th of month, otherwise to next month)

`USD x` should represent the product of `recipients x` multiplied by USD 32 (approximately the payout amount). Initially, this approximation should suffice, even though our payout amount is in SLE. We adjust it to USD 32 every few months.

Text for tile:
<img width="1026" alt="Screenshot 2024-01-30 at 00 52 54" src="https://github.com/socialincome-san/public/assets/6095849/7cb7a191-c778-4f62-9148-d064762e0aeb">

### Anything else?

There are currently two ways to make tables within a view in the admin tool. The standard way based on an existing FB collection (like [recipient list](https://staging-admin.socialincome.org/c/recipients)) or the more versatile version like in [Payments Confirmation](https://staging-admin.socialincome.org/payments-confirmation). I guess the second makes more sense here, but not a requirement. 

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

ssandino:
@sdwivedi I bring it up over lunch

PR #855, opened 2024-06-21
feature(admin): payout forecast

brennerthomas:
- Adding new collection which on loading calculates the six month forecast for payouts
- Conversion of 32 USD to SLE multiplied by number of potential participants
- solves [https://github.com/socialincome-san/public/issues/733](https://github.com/socialincome-san/public/issues/733)


ssandino (approved):

mkue (approved):
cool stuff! it seems to work well locally🎉 

I updated it to use the default payment amount that we have defined in the payment.ts file, so that we only have to change that in a single location if we update the payment amount again.

PR #870, opened 2024-07-21
[mobile feature]: single main.dart, app flavor from environment, less duplicated code

KarinBerg:
We had three main_*.dart files which contains almost the same code. So I refactored it to one again and read the app flavor from the environment.

TODO after merge: Adjust .vscode/launch.json


recipients_app/lib/main.dart
```diff
@@ -20,21 +20,32 @@ Future<void> main() async {
     [DeviceOrientation.portraitUp, DeviceOrientation.portraitDown],
   );
 
+  const appFlavor = String.fromEnvironment("FLUTTER_APP_FLAVOR");
```
KarinBerg, line 23: see https://docs.flutter.dev/deployment/flavors#retrieving-your-apps-flavor-at-runtime

ssandino (approved):

PR #883, opened 2024-08-06
feature(functions): support twint

andrashee:
The webhook importing twint charges throws an error:
```
Error Could not create user for Stripe customer: cus_QLt13VT23sUUFy, unknown id, email or name Error: Could not create user for Stripe customer: cus_QLt13VT23sUUFy, unknown id, email or name at StripeEventHandler.constructUser (/workspace/dist/shared/src/stripe/StripeEventHandler.js:157:23) at StripeEventHandler.getOrCreateFirestoreUser (/workspace/dist/shared/src/stripe/StripeEventHandler.js:54:43) at process.processTicksAndRejections (node:internal/process/task_queues:95:5) at async StripeEventHandler.storeCharge (/workspace/dist/shared/src/stripe/StripeEventHandler.js:176:29) at async StripeEventHandler.handleChargeEvent (/workspace/dist/shared/src/stripe/StripeEventHandler.js:28:24) at async /workspace/dist/functions/src/webhooks/stripe/index.js:41:41
```

The problem is, that we actively check if the user has a name field, which is not set for a twint payment.
https://github.com/socialincome-san/public/blob/77ac33184b22fc4562f6063953714dd34a1f5d16/shared/src/stripe/StripeEventHandler.ts#L183

```
{
  "id": "cus_Qbth7ZrV08LMcs",
  "object": "customer",
  "address": {
    "city": null,
    "country": null,
    "line1": null,
    "line2": null,
    "postal_code": null,
    "state": null
  },
  "balance": 0,
  "created": 1722921967,
  "currency": null,
  "default_source": null,
  "delinquent": false,
  "description": null,
  "discount": null,
  "email": "xyz@socialincome.org",
  "invoice_prefix": "12AD4FD6",
  "invoice_settings": {
    "custom_fields": null,
    "default_payment_method": null,
    "footer": null,
    "rendering_options": null
  },
  "livemode": false,
  "metadata": {},
  "name": null,
  "next_invoice_sequence": 1,
  "phone": null,
  "preferred_locales": [
    "en-GB"
  ],
  "shipping": null,
  "tax_exempt": "none",
  "test_clock": null
}
```

Since first and lastname are modelled as optional in our types, I assume we don't need them?

### DIFF
--- a/shared/src/stripe/StripeEventHandler.ts
+++ b/shared/src/stripe/StripeEventHandler.ts
@@ -180,8 +180,8 @@ export class StripeEventHandler {
 	 * This is mainly for failed payments where we didn't create a user through the website directly
 	 */
 	constructUser = (customer: Stripe.Customer): User => {
-		if (!customer.id || !customer.email || !customer.name) {
-			throw new Error(`Could not create user for Stripe customer: ${customer.id}, unknown id, email or name`);
+		if (!customer.id || !customer.email) {
+			throw new Error(`Could not create user for Stripe customer: ${customer.id}, unknown id, email`);
 		}
 		const { firstname, lastname } = splitName(customer.name);
 		return {
--- a/shared/src/types/user.ts
+++ b/shared/src/types/user.ts
@@ -50,7 +50,7 @@ export type User = {
 	created_at: Timestamp;
 };
 
-export const splitName = (name: string | undefined) => {
+export const splitName = (name: string | undefined | null) => {
 	if (!name) {
 		return {
 			firstname: undefined,
--- a/shared/src/types/user.ts
+++ b/shared/src/types/user.ts
@@ -50,7 +50,13 @@ export type User = {
 	created_at: Timestamp;
 };
 
-export const splitName = (name: string) => {
+export const splitName = (name: string | undefined) => {
+	if (!name) {
+		return {
+			firstname: undefined,
+			lastname: undefined,
+		};
+	}
 	const stripeNames = name.trim().split(' ');
 	const sanitizeName = (n: string) => n.charAt(0).toUpperCase() + n.slice(1).toLowerCase().trim();
 


shared/src/stripe/StripeEventHandler.ts
```diff
@@ -180,8 +180,8 @@ export class StripeEventHandler {
 	 * This is mainly for failed payments where we didn't create a user through the website directly
```
andrashee, line 180: @mkue, is this comment actually true? 
In the webhooks logs from Stripe, it seems that the user doesn't exist for a successful twint payment. Where do you create the user? Is it a race condition, whatever is faster?

shared/src/types/user.ts
```diff
@@ -50,7 +50,13 @@ export type User = {
 	created_at: Timestamp;
 };
 
-export const splitName = (name: string) => {
+export const splitName = (name: string | undefined) => {
+	if (!name) {
+		return {
+			firstname: undefined,
```
andrashee, line 56: To be safe, we could also use '' here instead.

ssandino:
> Since first and lastname are modelled as optional in our types, I assume we don't need them?

Yes, this is correct. Theoretically, there could be a contributor without a name. This ensures that if a user makes a payment (on stripe form) but never fills out "our" form afterwards, they are still handled as a normal user. I guess 99% will fill out the form after payment anyway.

shared/src/stripe/StripeEventHandler.ts
mkue: No, this comment is wrong. The usual way for a user to get created is through the Stripe webhook.

And yes, it's technically a race condition. If a new user fills out the CreateUserForm before the Stripe webhook has been called (which should only happen if Stripe has a problem), then the user is created when the form is submitted. This is a very unlikely case, though.

I don't see how this can be avoided as we need to store the user info on form submit.

PR #929, opened 2024-10-24
transparent bg on nav when hero video

mdumond:


mdumond:
yes we did not test at all :upside_down_face: I will check that this week end

mdumond:
I have just pushed a fix that fixes the nav background on scroll

ssandino:
works 🙌🏼

ssandino (approved):

mkue:
Very very cool🙌🏼 I'm super happy this is resolved now. It bothered me quite a bit that I didn't find a better solution initially🙃

I pushed one last commit that sets the default color to `null` in the `useNavbarBackgroundColor()` hook. This a) resolves an issue that the navbar stayed blue when navigating back to the home page, and b) makes it a bit more clear that this hook is there to overwrite the default background color (which is blue due to `theme-blue` that is used on the NavbarClient component).

mkue (approved):

PR #930, opened 2024-10-24
bugfix(website): display the correct logos in quotes section on home page

mdumond:


ssandino (approved):
🙌🏼

mdumond:
Some notes here: the worldbank logo is much stronger than the other 2, so either it needs to be a bit smaller, or the other 2 need to grow a bit. 

Also, I have run the svgs into https://svgomg.net/, it's a tool that allows to reduce the svg file sizes, though they were already quite good, they got reduced by ~10% each

I had to move the svgs to the /public folder, I did not remove them from (assets), maybe we should check that they are not used anywhere else and remove them from there so they are not doubled

mkue (approved):
nice! looks good to me :)

PR #944, opened 2024-11-17
Website: Hero video optimization (subtitles, flickering)

mdumond:
I replaced the call to an external file for the local translated versions. The files are correct in the html:

on live:
image

on the branch:
image

they don't appear yet though, but I also have trouble displaying the subtitles in the current main branch..

Also, the "label" argument is supposed to be the language name, I did not find in the code that we have a way to get that from the langcode, do we have something like that already?

## Summary by CodeRabbit

- **New Features**
	- Added support for video subtitles in multiple languages (German, English, French, Italian) to enhance accessibility.
	- Enhanced localization capabilities for the video component, allowing subtitles to adapt based on selected language and region.

- **Bug Fixes**
	- Improved error handling for video loading issues, ensuring a smoother user experience.

- **Documentation**
	- Updated component documentation to reflect new props and localization features.


ssandino:
@mathildedd @mkue we got the answer from the Mux support team who looked at this PR: 

Great to hear! So I see you're using the Mux Video element, the handling of captions would be under the vid element as you have. However, when I look and inspect that video element on your landing page, I'm only seeing the English subtitle in there:



Also, the Mux Video should attach all subtitles to the video element if you add the subtitles to your Mux Asset. You could then programmatically set the captions language with something like this:


```
const tracks = video.textTracks;

        // Function to activate Spanish subtitles
        function setSpanishSubtitle() {
                // Check if the track's language is Spanish (srclang="es")
                if (track.language === 'es') {
                    track.mode = 'showing'; // Display the track
                } else {
                    track.mode = 'disabled'; // Hide other tracks
                }
            }
        }
```

mkue (approved):
nice!

PR #959, opened 2024-11-30
chore(recipients-app): Bump android versions and dependencies

MDikkii:
So... I tried to run the app with Android Studio Ladybug installed and made the changes until the build worked again.

The one problem with this solution is that `Sentry` seems to be crashing at the start. That means that we may lose the sentry logs if it can't initialize properly. 

## Release Notes

- **New Features**
	- Enhanced localization support with new tooltip and date label methods.
  
- **Improvements**
	- Updated application namespace and ID for better identification.
	- Upgraded to Gradle 8.5 and updated plugin versions for improved compatibility.
	- Updated `intl` dependency to version 0.19.0 for potential new features and fixes.

- **Bug Fixes**
	- Adjusted button styling logic for consistent appearance across states. 

- **Chores**
	- Removed unnecessary text scaling from the MediaQuery widget in tests.

### DIFF
--- a/recipients_app/pubspec.yaml
+++ b/recipients_app/pubspec.yaml
@@ -25,7 +25,7 @@ dependencies:
     sdk: flutter
   flutter_native_splash: ^2.4.0
 
-  intl: ^0.18.0
+  intl: ^0.19.0
   intl_phone_number_input: ^0.7.4
   json_annotation: ^4.9.0
   json_serializable: ^6.8.0
--- a/recipients_app/test/helpers/golden_test_device_scenario.dart
+++ b/recipients_app/test/helpers/golden_test_device_scenario.dart
@@ -27,7 +27,6 @@ class GoldenTestDeviceScenario extends StatelessWidget {
             size: device.size,
             padding: device.safeArea,
             platformBrightness: device.brightness, 
-            textScaler: TextScaler.linear(device.textScaleFactor),
             devicePixelRatio: device.devicePixelRatio,
           ),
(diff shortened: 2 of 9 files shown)


recipients_app/test/helpers/golden_test_device_scenario.dart
```diff
@@ -27,7 +27,6 @@ class GoldenTestDeviceScenario extends StatelessWidget {
             size: device.size,
             padding: device.safeArea,
             platformBrightness: device.brightness, 
-            textScaler: TextScaler.linear(device.textScaleFactor),
```
KarinBerg, line 30: Why did you removed this line? Did you run the tests and then it failed and you had to remove this line?

recipients_app/pubspec.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
     sdk: flutter
   flutter_native_splash: ^2.4.0
 
-  intl: ^0.18.0
```
KarinBerg, line 28: I think you are using the wrong Flutter version. We are still on 3.19.6

KarinBerg (changes requested):
See Slack for why you had to do all this changes :)

recipients_app/test/helpers/golden_test_device_scenario.dart
MDikkii: No, new version of the dependency just show that there is no more TextScaler 🤔

PR #962, opened 2024-12-01
Website: Campaign form setup

andrashee:
Before doing the form itself, I want to quickly check with you @mkue, if the proposed architecture of the page fits into the overall one.
Idea:
- Static page /campaign to create a new one. Will contain the explanations, FAQ, ...).
- Client component for the form itself. Directly posts the data to firestore.

### DIFF
--- a/shared/locales/de/website-campaign.json
+++ b/shared/locales/de/website-campaign.json
@@ -1,4 +1,12 @@
 {
+  "metadata": {
+    "title": "Eigene Kampagne erstellen | Social Income"
+  },
+  "form": {
+    "title": "Erstelle deine Kampagne",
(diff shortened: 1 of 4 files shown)


mkue:
> Before doing the form itself, I want to quickly check with you @mkue, if the proposed architecture of the page fits into the overall one. Idea:
> 
> * Static page /campaign to create a new one. Will contain the explanations, FAQ, ...).
> * Client component for the form itself. Directly posts the data to firestore.

Yes, that definitely makes sense. As long as the form is just used on that one page, I would place the file direclty next to the page.tsx file that uses it.

I like that you used the `useMutation()` hook, I wanted to start using that also :)

Issue #761, opened 2024-02-29
[Website Feature]: Update survey background color


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

The current grey background used in our surveys does not align with the established color scheme of Social Income. 

![E3B90E65-475E-4CA5-9EB7-AD3964A0B138_1_101_o](https://github.com/socialincome-san/public/assets/6095849/529d3c3c-3413-4a80-a32d-2790e484999e)

Please note that the screenshot originates from an app that loads the React survey page as an iframe, which can also be directly accessed on the web. You find the survey code [here](https://github.com/socialincome-san/public/tree/main/website/src/app/%5Blang%5D/%5Bregion%5D/(survey-tool)/survey/%5Brecipient%5D/%5Bsurvey%5D).

### Describe the solution/feature

To resolve this issue, we propose either removing the grey background altogether or changing it to the standard blue color #EDF3FF that is consistently used across the Social Income.

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

devesh-2002:
May I take up this issue?

ssandino:
With pleasure @devesh-2002 – as the survey part is not yet documented in the readme, could you document the basic steps on your side (how to test, open etc. surveys) while working on it? This would help for future changes on the surveys.

devesh-2002:
Sure, I will try.

devesh-2002:
Is the survey page accessible only after Sign In? When I try to Sign In with Google, it does not sign in.

ssandino:
Can you write a short "hello" to hello@socialincome.org – then I add you to the staging admin tool.

devesh-2002:
Done. Please check it.

elizabeth-agada:
Is this issue still available for work? I'd love to contribute if it's not currently being worked on.

Issue #760, opened 2024-02-29
[Website Feature]: Disable cookie banner for surveys


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently, our surveys display a cookie banner that is originally intended for website visitors. This banner could potentially disrupt the user experience for recipients completing the surveys, particularly those who are speakers of Krio.

![08B4D98A-E42C-478A-8D43-86A43170AD0D_1_101_o](https://github.com/socialincome-san/public/assets/6095849/97e484ca-5520-4e7b-be38-2bc720f5b403)

### Describe the solution/feature

The cookie banner should either be completely removed or made inactive within the survey environment.

### Anything else?

While implementing this change, it's crucial to ensure that the main website remains unaffected, as it uses the same cookie banner component for various functionalities.

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

irolinski:
I'd like to work on this. However, I am also having trouble logging in (using sign in with Google).

Issue #759, opened 2024-02-29
[Website Feature]: Remove region selection dropdown for surveys


ssandino:
### Is there an existing request for this feature?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

In our surveys, we present an option for users to select their region via a dropdown. While the dropdown is not pre-selected (like on the our website), it also only lists `international` (and not `ch`). Regardless, a dropdown to choose region is not needed for the survey as the survey content remains identical across all regions.

Attached is a screenshot illustrating the survey in its current state, displaying the popup and the unnecessary dropdown. Although these screenshots are from the app, the survey is actually presented within an iframe, and it's often also accessed by recipients directly through a URL (for the ones without the app). Therefore, this issue is not specific to the app and a website issue. 

![1ABC6E95-E30F-4205-A20B-A9664300961E_1_101_o](https://github.com/socialincome-san/public/assets/6095849/4d7fb7ec-db56-47b7-b6ad-675f2be1c59d)

### Describe the solution/feature

The optimal solution is to remove the region selection dropdown from the survey interface. While implementing this change, it's crucial to ensure that the [main website](socialincome.org) remains unaffected, as it might utilize the same component.

### Anything else?

It is important to note that we should still provide the option to select the survey language, as we offer the surveys in both English and Krio.

### Code of Conduct

- [X] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

txpham209:
I would like to work on this issue.

elizabeth-agada:
Is this issue still available for work? I'd love to contribute if it's not currently being worked on.

ssandino:
@elizabeth-agada This issue is challenging as it requires some pre-work and backend knowledge, given that surveys are protected and only accessible to recipients, making it hard to simulate. Can we propose a different issue?

elizabeth-agada:
Oh, okay
Yeah, sure, I can take on another issue.
And sorry for late response.

ssandino:
@elizabeth-agada what about this #728?

PR #971, opened 2024-12-06
Website: Survey page (fixing bugs)

dnhn:
Hi team,

I found this project through [For Good First Issue](https://forgoodfirstissue.github.com). After reviewing issues #759, #760, #761, I noticed they are all related to the Survey page and have remained unresolved since their last activities. I decided to investigate and fix them.

Let me know if you have any feedback!

## Release Notes

- **New Features**
	- Introduced a custom hook `useIsPage` to determine if the current page is a survey page, enhancing component logic.
	- Conditional rendering for regions dropdown in the `I18nDialog` component based on the current page type.
	- Updated logic in the `CookieConsentBanner` to hide the banner on survey pages.

- **Style**
	- Updated the `--background-dim` CSS variable to `transparent` for improved visual clarity on survey pages.

### DIFF
--- a/website/src/hooks/useIsPage.ts
+++ b/website/src/hooks/useIsPage.ts
@@ -2,7 +2,11 @@ import { usePathname } from 'next/navigation';
 
 export function useIsPage(page: string) {
 	const pathname = usePathname();
+	// URL structure: /[lang]/[region]/[page]
+	const segments = pathname.split('/');
 
-	// /[lang]/[region]/[page]
-	return pathname.split('/')[3] === page;
+	// If `page` is blank and `segments` length is less than 4, it’s the home page
+	if (!page) return segments.length < 4;
+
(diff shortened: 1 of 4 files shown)


website/src/hooks/useIsPage.ts
```diff
@@ -0,0 +1,12 @@
+import { usePathname } from 'next/navigation';
+
+export function useIsPage(page: string) {
```
mkue, line 3: This hook seems to have a quite specific use-case. It only checks for the third item in the path segments. Maybe a hook `usePathnameContains(page: string)` would be more re-usable?

You then still do 
```ts
const isSurveyPage = usePathnameContains('/[lang]/[region]/survey');
```

What do you think?

dnhn: That’s a good idea, I’ll update it.

PR #973, opened 2024-12-06
Website: Home page explainer video (fix bug)

dnhn:
## Issue
The video has a fixed `width` (`w-96 = 384px`) and breaks the layout when viewing in mobile screens.

## Solution
Setting the video’s `max-width` to 100% to prevent overflow regardless of the fixed width.

## Summary by CodeRabbit

- **New Features**
	- Improved layout flexibility for the explainer video component, allowing it to scale responsively within its parent container.


ssandino:
Nice catch. Fixed!

mkue (approved):
nice 🙏

PR #986, opened 2024-12-16
Website: User location based flag

### LINKED ISSUE #978: [Web]: Update flags to reflect user location, not currency
### Is your feature request related to a problem? Please describe.

Currently, the flag in the [navigation](https://socialincome.org/en/int) is linked to the currency, which can be confusing since flags are typically associated with locations. Additionally, displaying the user’s own country flag helps build trust.

Current Status:
Flag linked to currency. Since we only support three currencies, there are currently only three flags displayed.

Image

### Describe the solution/feature

The goal is to link the flag to the user’s location, which is determined in the [middleware](https://github.com/socialincome-san/public/blob/main/website/src/middleware.ts). Based on this information, the appropriate country flag will be displayed. 

This approach will remain independent of the current setup where all users outside Switzerland are routed to the International page (.../int/...). This routing behavior should not change.

### Describe alternatives you've considered

_No response_

### Criteria for Completion

_No response_

### Anything else?

_No response_

### Code of Conduct

- [x] I've read the Code of Conduct and understand my responsibilities as a member of the Social Income community

dnhn:
Using `geolocation` from `@vercel/functions` to get information based on user request.
https://vercel.com/docs/functions/vercel-functions-package#geolocation

Resolve #978

## Summary by CodeRabbit

- **New Features**
  - Introduced a new API route for geolocation data retrieval.
  - Added a custom hook for fetching geolocation data.
  - Enhanced the navbar to display country flags based on geolocation.
  - Added a function to construct URLs for flag images based on country names.
  - Introduced a constant for managing country cookies.

- **Improvements**
  - Updated the `DefaultParams` interface to include an optional `country` property.
  - Enhanced validation for country codes within the middleware.

- **Dependency Updates**
  - Added a new dependency for Vercel functions.

- **Bug Fixes**
  - Replaced local flag image URL construction with a centralized utility function.


ssandino:
Great job @dnhn. I am testing three different scenarios: 

###  1. URL without language or location (socialincome.org)
Test URL: https://public-git-nhan-user-location-flag-social-income.vercel.app on [TestedLocally](https://testlocal.ly/tests/676016de0595c298780cdf4e)

Result: ✅ shows location specific flags

Screenshot 2024-12-16 at 13 03 20

###  2. URL **with** language and "int" location (socialincome.org/de/int) 
Test URL: https://public-git-nhan-user-location-flag-social-income.vercel.app/de/int on [TestedLocally](https://testlocal.ly/tests/6760192ab0ac4cc0800b2740)

Result: ✅ shows location specific flags

Screenshot 2024-12-16 at 13 13 05

###  3. URL **with** language and location, for which we have a specific website setup (only Switzerland for now = ch) (socialincome.org/de/ch) 
Test URL: https://public-git-nhan-user-location-flag-social-income.vercel.app/de/ch on [TestedLocally](https://testlocal.ly/tests/676017675b9f37230501f5b6)

Test result ❌ The international flags are displayed, but the Swiss flag should be shown in this case, as the page is routed to a specific country page which we serve (in this case /ch/). For all other cases see Nr. 2 above, which is fine.

Screenshot 2024-12-16 at 13 14 16

PS. Not sure how good this [TestedLocally](https://testlocal.ly) site is for testing this. But didn't have a VPN installed to test otherwise.

dnhn:
> Test result ❌ The international flags are displayed, but the Swiss flag should be shown in this case, as the page is routed to a specific country page which we serve (in this case /ch/). For all other cases see Nr. 2 above, which is fine.

@ssandino the currency flag is displayed only if the region is different from international, isn’t it?

> PS. Not sure how good this [TestedLocally](https://testlocal.ly) site is for testing this. But didn't have a VPN installed to test otherwise.

This looks good enough to me 👍🏼

ssandino:
So the logic for the location flag should be as follows:

- If the URL contains …/language/CH, always show the Swiss flag (regardless of what you get from the newly introduce geo function).
- If the URL contains nothing or …/language/int, show what you get from the geo function.

Currently, we only have a Swiss-specific site, and the “int” version serves as a catch-all for other regions. However, in the future, we will have additional country-specific pages that should function similarly to the “ch” case.

dnhn:
> So the logic for the location flag should be as follows:
> 
> * If the URL contains …/language/CH, always show the Swiss flag (regardless of what you get from the newly introduce geo function).
> * If the URL contains nothing or …/language/int, show what you get from the geo function.
> 
> Currently, we only have a Swiss-specific site, and the “int” version serves as a catch-all for other regions. However, in the future, we will have additional country-specific pages that should function similarly to the “ch” case.

Does this means that the currency flag is always Swiss even if we select other currencies?

image

website/src/components/navbar/navbar-client.tsx
```diff
@@ -211,7 +221,17 @@ const MobileNavigation = ({ lang, region, languages, regions, currencies, naviga
 							{translations.myProfile}
 						</NavbarLink>
 						<div className="flex-inline flex items-center">
-							{Flag && <Flag className="mx-3 h-6 w-6 rounded-full" />}
+							{country && (
+								<Image
```
mkue, line 225: Should we also do it like this in `website/src/components/ui/currency-selector.tsx`? Then we could uninstall the package. Wdyt?

website/src/app/[lang]/[region]/(website)/me/hooks.ts
```diff
@@ -170,3 +171,21 @@ export const useAddEmployer = () => {
 		await queryClient.invalidateQueries({ queryKey: ['me', 'employers'] });
 	};
 };
+
+export const useGeolocation = () => {
```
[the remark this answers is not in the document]
Comment (dnhn) [reply website/src/app/[lang]/[region]/(website)/me/hooks.ts:175]: I had the same thought but then found the query hooks here. So I'll move this query first and we will migrate the rest later.

website/src/components/navbar/navbar-client.tsx
dnhn: Sure, I’ll update it.

dnhn: The URL we are using `https://raw.githubusercontent.com/lipis/flag-icons/a87d8b256743c9b0df05f20de2c76a7975119045/flags/…` only accepts country code, do you have any library suggestions for currency flags?

I would work on this on another PR since this involves with other components.

mkue: Do you mean currency icons? Otherwise, we would have to create some mapping to map currencies to countries (similiar to the `countryToCurrency` mapping in `currency.ts`).

FYI – I created an MR (#992) to use a middleware to set the country code. Let me know what you think :)

mkue (approved):

PR #992, opened 2024-12-22
Set cookie with location of request country

mkue:
Hey @dnhn! Looking at your #986 MR again, I realized that we can already set the country code of the request location as a cookie in the middleware. Doing it like this, it's not necessary anymore to do an extra request to the API endpoint you created because you can directly access the info in the cookie.
What do you think about this approach?

Seems to work as well:

Screenshot 2024-12-22 at 16 14 00

### DIFF
--- a/shared/src/types/country.ts
+++ b/shared/src/types/country.ts
@@ -250,3 +250,5 @@ export const COUNTRY_CODES = [
 	'ZW', // 'Zimbabwe',
 ] as const;
 export type CountryCode = (typeof COUNTRY_CODES)[number];
+
+export const isValidCountryCode = (code: string): code is CountryCode => COUNTRY_CODES.includes(code as CountryCode);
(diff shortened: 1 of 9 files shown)


dnhn (changes requested):
Hey @mkue, this is a great idea 💡 I didn’t think of the middleware approach!

However, the flag should follow the region when it’s not `/int` (which we have `/ch` for now). That’s why I checked it using `isIntRegion = region === 'int'`, can we bring that back or you can suggest a better way.

image

ssandino (approved):
Thanks. Yes we should focus on location (and ignore currency for the flag). /ch should always show Swiss flag. /int should always show actual location

mkue:
> Hey @mkue, this is a great idea 💡 I didn’t think of the middleware approach!
> 
> However, the flag should follow the region when it’s not `/int` (which we have `/ch` for now). That’s why I checked it using `isIntRegion = region === 'int'`, can we bring that back or you can suggest a better way.

Yes, that makes total sense. Sorry, I should not have removed that. I put it back [here](https://github.com/socialincome-san/public/pull/992/commits/9615631ae0fe3ee84e25135fa01565f013e1e679).

mkue:
I'll approve the other MR. Feel free to merge it once this is merged into your branch. All looks good to me. Thanks a lot for your help!

dnhn (approved):

PR #1017, opened 2025-01-22
new footer design

mdumond:
## Summary by CodeRabbit

- **UI/UX Improvements**
  - Enhanced footer text prominence by updating link sizes.
  - Simplified footer link presentation by removing icons.
  - Adjusted footer layout and padding for improved design.
  - Updated footer's responsive design for better adaptability across screen sizes.
  - Modified styling approach in the BaseContainer for a more streamlined design using `baseClassNames`.


mdumond:
I forgot to add here that I did not manage to get the line height from Figma considering the font-size, because as far as I could tell they were linked together in the code.

ssandino:
Looking good. Don't worry about line height, it's not pixel perfect on Figma either. It's even better to start with the default. And only make exception when really needed.

mkue (approved):
nice! thank you :)

I added a little change that you can directly set the theme on the base class container. I don't think we need the restriction to just allow setting the background color.

mdumond:
oh nice thanks! I was a bit unhappy with the background overrides

PR #1024, opened 2025-01-24
Add linksCn (links className helper) to UI-Kit

almsh:


mkue (approved):
nice! I updated the className in the Links that we already used now 👍

PR #1067, opened 2025-03-08
Website: Remove conditional display of partners, Fix permalink present in partner card

DarkMenacer:
## Summary by CodeRabbit

- **New Features**
  - Introduced a constant for NGO identifiers, enhancing the organization of NGO data.
  
- **Bug Fixes**
  - Improved the display of recipient statistics by providing default values when data is missing, ensuring that badges consistently show accurate information for all organizations.
  
- **Enhancements**
  - Updated NGO translation functionality to handle cases where an NGO may not be found, improving error handling.


ssandino:
@mkue could you also include this in the release, as it fixes a bug that suddenly not all partners are mentioned on the partner page (and they noticed)

website/src/app/[lang]/[region]/(website)/partners/[slug]/page.tsx
```diff
@@ -3,9 +3,17 @@ import { PartnerHome } from '@/app/[lang]/[region]/(website)/partners/(component
 import { NgoEntryJSON } from '@/app/[lang]/[region]/(website)/partners/(types)/PartnerCards';
 import { Translator } from '@socialincome/shared/src/utils/i18n';
 import { redirect } from 'next/navigation';
+import { ngos } from '../(sections)/ngolist';
 
-function getNGOTranslation(translator: Translator, slug: string): NgoEntryJSON {
-	return translator.t(slug);
+function getNGOTranslation(translator: Translator, slug: string): NgoEntryJSON | undefined {
```
Comment (mkue) [inline website/src/app/[lang]/[region]/(website)/partners/[slug]/page.tsx:8]: This works, but is a bit weird imo. It would be much cleaner in my opinion to have all ngos in the translation file in one object and then translate that entire object. The object with the translations of all ngs could then be passed to the PartnerHome component.

mkue (approved):
It's ok the way it is, but it could be a bit cleaner if you would only have one translate.t() call (of an object will all ngos) and display that in the PartnerHome component (obviously, the translation keys have to match then).

No need to do this in this MR though :)

ssandino:
@mkue : merged it with @DarkMenacer’s OK so we can fix the issue swiftly. He’ll handle the remaining points in a future PR.

PR #1137, opened 2025-06-15
[Mobile App] Use new Twilio phone verification service

KarinBerg:
This PR replaces the Firebase Auth phone verification with the new one from Twilio. 
(Needed Google Function is already implemented and deployed. See function '[verifyOtpFunction](https://github.com/socialincome-san/public/blob/main/functions/src/functions/webhooks/auth/twilio-verify.ts)')

- I introduced a new Auth Service with two specific implementations. One which uses Twilio and one which uses the old  Firebase Auth implementation.
- This new Auth Service is responsible for the phone number verification of the user, sign-in and sign-out of the user.
- UserRepository is now only responsible for fetching and updating a recipient

**Things to test:**
- Demo Mode and the switch between normal mode and demo mode
- User sign-up (User does not exists in Firebase Auth, you can delete your auth user in Firebase Console)
- Test Resend of OTP code
- User sign-in (User exists in Firebase Auth)
- User sign-out

## Summary by CodeRabbit

- **New Features**
  - Integrated Twilio-based OTP authentication alongside Firebase OTP, enabling phone number verification via Twilio.
  - Added support for Twilio credentials and configuration in build and launch settings.

- **Refactor**
  - Updated authentication logic to use a new `AuthService` abstraction, replacing direct repository calls in authentication flows.
  - Simplified user data source interfaces by removing authentication-related methods from user data management.

- **Bug Fixes**
  - Enhanced error localization to support Firebase Functions exceptions for clearer user messages.

- **Chores**
  - Added new dependencies for Twilio and Firebase Cloud Functions integration.


recipients_app/lib/my_app.dart
```diff
@@ -38,6 +39,8 @@ class MyApp extends StatelessWidget {
   final OrganizationRemoteDataSource organizationRemoteDataSource;
   final OrganizationDemoDataSource organizationDemoDataSource;
 
+  final AuthService authService;
+
```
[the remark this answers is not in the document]
KarinBerg: @novas1r1 Is that right?

recipients_app/lib/data/services/auth_service.dart
```diff
@@ -0,0 +1,100 @@
...
+  Future<void> submitVerificationCode(String code);
+
+  Future<void> signInWithCredential();
+
+  @nonVirtual
+  Stream<User?> authStateChanges() {
+    final StreamController<User?> authStateController = StreamController();
+    StreamSubscription<User?>? authStateSubscription;
+
+    authStateSubscription = _getInternalAuthStateSteam(false).listen((authState) {
```
KarinBerg, line 34: @novas1r1 I'm not happy with this "demo mode auth stream implementation" but I do not know a better solution. Any suggestions?

recipients_app/lib/data/services/auth_service.dart
```diff
@@ -0,0 +1,89 @@
...
+    required this.demoManager,
+  });
+
+  Future<void> verifyPhoneNumber({
+    required String phoneNumber,
+    required Function() onCodeSend,
+    required Function(AuthException) onVerificationFailed,
+    required Function() onVerificationCompleted,
+  });
+
```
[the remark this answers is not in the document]
KarinBerg: Please explain more

KarinBerg:
@coderabbitai full review

ssandino (approved):

recipients_app/lib/data/services/auth_service.dart
novas1r1: @MDikkii implemented the demo mode so he can maybe give us more details here. But I will check if theres a better option. Why exactly are you not happy with it?

recipients_app/lib/data/services/auth_service.dart
```diff
@@ -0,0 +1,97 @@
...
+    await firebaseAuth.signOut();
+    demoManager.isDemoEnabled = false;
+  }
+
+  void dispose() {
+    _demoUserStreamController.close();
+  }
+}
+
+class AuthException implements Exception {
```
novas1r1, line 91: In the future we need to catch different auth exceptions and maybe give valuable and translated messages to the users. Currently if an error occurs I'm getting "AuthException: Verification service not initialized" for example which is not helpful at all to a user :D. I think this belongs to a separate ticket though because it affects the whole app.

recipients_app/lib/view/error_localization_helper.dart
```diff
@@ -1,7 +1,19 @@
+import "package:cloud_functions/cloud_functions.dart";
 import "package:firebase_auth/firebase_auth.dart";
 import "package:flutter_gen/gen_l10n/app_localizations.dart";
 
 String localizeExceptionMessage(Exception? ex, AppLocalizations localizations) {
+  if (ex is FirebaseFunctionsException) {
```
novas1r1, line 6: okay I forgot we added propper error handling to firebase exceptions already. maybe we can just add the auth exceptions here

novas1r1 (approved):

PR #1258, opened 2025-09-05
Mobile: Update to new dataschema

novas1r1:
THIS NEEDS TESTING!

- Migrated to new dataschema.
- Updated Flutter to 3.38.4 due to issues

### DIFF
--- a/recipients_app/lib/l10n/arb/app_en.arb
+++ b/recipients_app/lib/l10n/arb/app_en.arb
@@ -52,7 +52,6 @@
     "nextPayment": "Next Payment",
     "review": "Review",
     "payments": "Payments",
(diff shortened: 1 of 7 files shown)


recipients_app/pubspec.yaml
```diff
@@ -2,58 +2,54 @@ name: app
 description: Social Income App
 
 publish_to: "none"
-version: 1.1.17+1 # The build number is set in Codemagic automatically.
+version: 1.2.0+1 # The build number is set in Codemagic automatically.
```
KarinBerg, line 5: Maybe we should increase the version to 2.0.0 because it is a breaking change for the user and for us it is clear to remember that we use since 2.0 the new API.

PR #1270, opened 2025-10-07
feat: adapt all services to new db schema

rawi96:


shared/src/database/schema.prisma
```diff
@@ -9,394 +9,469 @@ datasource db {
...
+  endDate                DateTime       @map("end_date")
+  isActive               Boolean        @default(false) @map("is_active")
+  public                 Boolean?       @map("public")
+  featured               Boolean?       @map("featured")
+  slug                   String?        @map("slug")
+  metadataDescription    String?        @map("metadata_description")
+  metadataOgImage        String?        @map("metadata_og_image")
+  metadataTwitterImage   String?        @map("metadata_twitter_image")
+  creatorName            String?        @map("creator_name")
+  creatorEmail           String?        @map("creator_email")
```
mkue: we imo need the organizationId here too

PR #1303, opened 2025-10-22
add organization switcher and cleanup views

rawi96:
## Summary by CodeRabbit

* **New Features**
  * Active organization selector in Account Settings and server-side action to save it
  * Organization Members and Program Members pages
  * Programs dropdown in the navigation and “Create new program” quick action
  * User Programs dashboard with editable vs view-only program cards

* **Bug Fixes & Improvements**
  * Campaigns now filtered by active organization
  * Simplified navigation and consolidated management flows; removed legacy scoped pages

* **Refactor**
  * Streamlined member and permission views across organizations and programs

### DIFF
--- a/website/src/app/portal/components/navbar/navbar-desktop.tsx
+++ b/website/src/app/portal/components/navbar/navbar-desktop.tsx
@@ -7,7 +7,7 @@ import {
 	DropdownMenuContent,
 	DropdownMenuItem,
 	DropdownMenuSeparator,
-	DropdownMenuTrigger
+	DropdownMenuTrigger,
 } from '@/app/portal/components/dropdown-menu';
 import { Logo } from '@/app/portal/components/logo';
 import { useNavbarLinks } from '@/app/portal/components/navbar/hooks/use-navbar-links';


website/src/app/portal/components/navbar/hooks/use-navbar-links.tsx
```diff
@@ -5,27 +5,28 @@ export type NavLink = {
 	href: string;
 	label: string;
 	activeBase?: string;
```
fabricem90, line 7: is `activeBase` still being used?

website/src/app/portal/components/navbar/hooks/use-navbar-links.tsx
```diff
@@ -39,19 +40,14 @@ export const useNavbarLinks = (user: UserInformation) => {
...
 			icon: Settings,
 		},
 	];
 
-	const isActiveLink = (path: string, href: string, exact?: boolean, activeBase?: string) => {
-		if (exact) return path === href;
-		const base = activeBase ?? href;
-		return path === base || path.startsWith(base + '/');
-	};
+	const isActiveLink = (path: string, href: string) => path.startsWith(href);
```
fabricem90: I think this only works with the first tab/subroute in any main nav route. In case there's always only one sub-route per main nav route, we could get the base path per main nav with something like `const basePath = href.substring(0, href.lastIndexOf("/"));` and `path.startsWith(basePath)`

PR #1304, opened 2025-10-27
feat: add recipient form for editing and creating new recipients

fabricem90:
## Summary by CodeRabbit

* **New Features**
  * Recipient create/update/get server actions and options endpoint added.
  * Program and Local Partner dropdown option endpoints added.
  * New Switch UI component for boolean fields.

* **Improvements**
  * Recipient form converted to a dynamic, schema-driven form with async option loading, unified submit flow, and onError handling.
  * Contact fields accept nullable/empty values and include a Has WhatsApp flag.
  * Dynamic form: improved enum, accordion, and validation handling.

* **UI**
  * Table row opens recipient edit dialog; table component now receives userId prop.

### DIFF
--- a/shared/src/database/services/local-partner/local-partner.service.ts
+++ b/shared/src/database/services/local-partner/local-partner.service.ts
@@ -3,8 +3,8 @@ import { BaseService } from '../core/base.service';
 import { ServiceResult } from '../core/base.types';
 import {
 	LocalPartnerCreateInput,
+	LocalPartnerOption,
 	LocalPartnerPayload,
-	LocalPartnerRecipientOption,
 	LocalPartnerTableView,
 	LocalPartnerTableViewRow,
 	LocalPartnerUpdateInput,
@@ -67,7 +67,7 @@ export class LocalPartnerService extends BaseService {
 		}
 	}
 
-	async getRecipientOptions(userId: string): Promise<ServiceResult<LocalPartnerRecipientOption[]>> {
+	async getOptions(userId: string): Promise<ServiceResult<LocalPartnerOption[]>> {
 		try {
 			const user = await this.db.user.findUnique({
 				where: { id: userId },
--- a/shared/src/database/services/local-partner/local-partner.types.ts
+++ b/shared/src/database/services/local-partner/local-partner.types.ts
@@ -30,7 +30,7 @@ export type LocalPartnerPayload = {
 	};
 };
 
-export type LocalPartnerRecipientOption = {
+export type LocalPartnerOption = {
 	id: string;
 	name: string;
 };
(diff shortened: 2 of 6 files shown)


shared/src/database/services/local-partner/local-partner.service.ts
```diff
@@ -66,6 +67,30 @@ export class LocalPartnerService extends BaseService {
 		}
 	}
 
+	async getRecipientOptions(userId: string): Promise<ServiceResult<LocalPartnerRecipientOption[]>> {
```
rawi96: Could we rename this to something like just getAll()?

because it has actually nothing todo with recipients.

shared/src/database/services/local-partner/local-partner.types.ts
```diff
@@ -30,5 +30,10 @@ export type LocalPartnerPayload = {
 	};
 };
 
+export type LocalPartnerRecipientOption = {
```
rawi96: same here

maybe just export type = LocalPartner = {

shared/src/database/services/program/program.service.ts
```diff
@@ -1,9 +1,33 @@
 import { PayoutStatus, ProgramPermission } from '@prisma/client';
 import { BaseService } from '../core/base.service';
 import { ServiceResult } from '../core/base.types';
-import { ProgramWallet, ProgramWalletView } from './program.types';
+import { ProgramRecipientOption, ProgramWallet, ProgramWalletView } from './program.types';
 
 export class ProgramService extends BaseService {
+	async getRecipientPrograms(userId: string): Promise<ServiceResult<ProgramRecipientOption[]>> {
```
rawi96: here the same
We could just name it "getByUserId"

shared/src/database/services/program/program.types.ts
```diff
@@ -13,3 +13,8 @@ export type ProgramWallet = {
 export type ProgramWalletView = {
 	wallets: ProgramWallet[];
 };
+
+export type ProgramRecipientOption = {
```
rawi96: same here

website/src/components/dynamic-form/dynamic-form.tsx
```diff
@@ -288,9 +292,10 @@ const GenericFormField = ({
 										</SelectTrigger>
 									</FormControl>
 									<SelectContent {...form.register(optionKey)}>
-										{Object.keys(getEnumValues(option, parentOption)).map((key) => (
-											<SelectItem value={key} key={key}>
-												{key}
+										{Object.entries(getEnumValues(option, parentOption)).map(([label, id]) => (
+											// TODO:
```
rawi96: todo what?

shared/src/database/services/local-partner/local-partner.service.ts
fabricem90: true, but it currently only returns the needed props for the recipient forms, and not all data which exists for local partner. Should we name it to something like `getList` ?

fabricem90: or return all the data available for local partner so that i could be re-used for other purposes

PR #1344, opened 2025-11-04
feat: check edit permissions for recipient creation and update

fabricem90:
## Summary by CodeRabbit

* **New Features**
  * Per-row permission included in recipients table and a top-level readOnly prop to control view vs edit behavior.
  * Date picker now accepts a customizable placeholder (defaults to "Select date").

* **Bug Fixes**
  * Stricter permission checks for creating, viewing, and updating recipients to block unauthorized edits.
  * Dialog titles now use "View" in read-only mode; form fields render "-" when read-only.

### DIFF
--- a/firestore_to_postgres/src/recipient/recipient.transformer.ts
+++ b/firestore_to_postgres/src/recipient/recipient.transformer.ts
@@ -16,34 +16,55 @@ export class RecipientTransformer extends BaseTransformer<FirestoreRecipientWith
 
 			const legacyPartnerId = raw.organisation?.id.split('/').pop();
 
-			const communicationPhone = raw.communication_mobile_phone?.phone
-				? {
-						connectOrCreate: {
-							where: { number: this.mapPhoneNumber(raw.communication_mobile_phone.phone) },
-							create: {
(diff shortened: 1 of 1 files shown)


website/src/app/portal/components/data-table/clients/recipients-table-client.tsx
```diff
@@ -21,19 +22,16 @@ export function RecipientsTableClient({
 
 	const [recipientId, setRecipientId] = useState<string | undefined>();
 	const [hasError, setHasError] = useState(false);
-
-	const [readOnly, setReadOnly] = useState(false);
+	const readOnly = rows.some((row) => row.permission === ProgramPermission.readonly);
```
rawi96: I’m not fully convinced this is the best approach.

Instead, we could pass a readonly flag into the tableClient, so we only need to decide edit permissions in one place:
	•	On a program-scoped page, we already know whether the user can edit that specific program.
Since page.tsx calls getAuthenticatedUserOrRedirect, which eventually calls UserService.getCurrentUserInformation, we could extend that method to include the required permission info.
	•	On the monitoring page, the button should only be enabled if the user has edit access to at least one program.

PR #1364, opened 2025-11-12
feat: create donation certificate

fabricem90:
## Summary by CodeRabbit

* **New Features**
  * Donation certificate generation: management UI to select contributors, year and language, batch-generate PDFs, and store/download certificates.
  * Enhanced multi-select component in the portal with improved usability, animations, and accessibility.

* **Chores**
  * Switched API authentication to use scheduler credentials.
  * Removed the email templating system and related templates/dependencies.
  * Updated localization strings for donation certificates in multiple languages.

### DIFF
--- a/shared/src/database/services/contribution/contribution.service.ts
+++ b/shared/src/database/services/contribution/contribution.service.ts
@@ -1,4 +1,5 @@
 import { Contribution } from '@prisma/client';
+import { endOfYear, startOfYear } from 'date-fns';
 import { BaseService } from '../core/base.service';
 import { ServiceResult } from '../core/base.types';
 import { OrganizationAccessService } from '../organization-access/organization-access.service';
@@ -168,13 +169,13 @@ export class ContributionService extends BaseService {
 		year: number,
 	): Promise<ServiceResult<ContributionDonationEntry[]>> {
 		try {
+			const start = startOfYear(new Date(year, 0, 1));
+			const end = endOfYear(new Date(year, 0, 1));
+
 			const result = await this.db.contribution.findMany({
 				where: {
 					contributorId: { in: contributorsIds },
-					AND: [
-						{ createdAt: { gte: new Date(`${year}-01-01 00:00:00`) } },
-						{ createdAt: { lte: new Date(`${year}-12-31 23:59:59`) } },
-					],
+					AND: [{ createdAt: { gte: start } }, { createdAt: { lte: end } }],
 				},
 				select: {
 					contributorId: true,
--- a/shared/src/database/services/contributor/contributor.service.ts
+++ b/shared/src/database/services/contributor/contributor.service.ts
@@ -203,11 +203,7 @@ export class ContributorService extends BaseService {
 	async getByIds(contributorIds?: string[]): Promise<ServiceResult<ContributorDonationCertificate[]>> {
 		try {
 			const result = await this.db.contributor.findMany({
-				where: contributorIds
-					? {
-							id: { in: contributorIds },
-						}
-					: undefined,
+				where: contributorIds && contributorIds.length > 0 ? { id: { in: contributorIds } } : {},
 				select: {
 					id: true,
 					account: true,
--- a/shared/src/database/services/donation-certificate/donation-certificate-writer.ts
+++ b/shared/src/database/services/donation-certificate/donation-certificate-writer.ts
@@ -20,7 +20,7 @@ export class DonationCertificateWriter {
 		this.contributions = contributions;
 	}
 
-	private getContributionsByCurrency = async (contributions: ContributionDonationEntry[], year: number) => {
+	private groupContributionsByCurrency = async (contributions: ContributionDonationEntry[], year: number) => {
 		return _(contributions)
 			.filter((contribution) => contribution.createdAt.getFullYear() === year)
 			.groupBy((contribution) => contribution.currency)
@@ -32,7 +32,7 @@ export class DonationCertificateWriter {
 	 * The function terminates when the PDF file has been written to the file at the given path.
 	 */
 	writeDonationCertificatePDF = async (filePath: string, language: LanguageCode) => {
-		const contributionsByCurrency = await this.getContributionsByCurrency(this.contributions, this.year);
+		const contributionsByCurrency = await this.groupContributionsByCurrency(this.contributions, this.year);
 
 		const translator = await Translator.getInstance({
 			language,


shared/src/database/services/contributor/contributor.service.ts
```diff
@@ -199,6 +200,52 @@ export class ContributorService extends BaseService {
 		}
 	}
 
+	async getForDonationCertificate(contributorIds?: string[]): Promise<ServiceResult<ContributorDonationCertificate[]>> {
```
rawi96: could we rename this to smth like getByIds?

shared/src/database/services/donation-certificate/donation-certificate.service.ts
```diff
@@ -62,4 +71,88 @@ export class DonationCertificateService extends BaseService {
...
+						usersSkipped.push(contributor.id);
+						return;
+					}
+
+					const existingCertificate = await this.db.donationCertificate.findMany({
+						where: {
+							year: year,
+							contributorId: contributor.id,
+						},
+					});
```
rawi96: pls extract this in a new private method like eg "findByYearAndContributor"

shared/src/database/services/donation-certificate/donation-certificate.service.ts
```diff
@@ -62,4 +71,88 @@ export class DonationCertificateService extends BaseService {
...
+
+					if (existingCertificate.length) {
+						console.info(
+							`User ${contributor.id} already has a certificate for year ${year}, skipping donation certificate creation`,
+						);
+						usersSkipped.push(contributor.id);
+						return;
+					}
+
+					let contributions = await this.contributionService.getForContributor(contributor.id);
```
rawi96: we should avoid to do db queries within a loop

if we e.g. generate certs for ALL 3800 contributors this will end in 3800 db queries. (same for the db query above with the existingCertificates)

shared/src/database/services/donation-certificate/donation-certificate.service.ts
```diff
@@ -62,4 +71,88 @@ export class DonationCertificateService extends BaseService {
...
+							data: {
+								year: year,
+								storagePath: destinationFilePath,
+								contributor: {
+									connect: {
+										id: contributor.id,
+									},
+								},
+							},
+						});
```
rawi96: Also here. Consider to gather all data first and then use db.createMany() instead of calling create() within a loop

shared/src/database/services/donation-certificate/donation-certificate-writer.ts
```diff
@@ -0,0 +1,162 @@
...
+import * as path from 'path';
+import PDFDocument from 'pdfkit';
+import { LANGUAGE_CODES } from '../../../types/language';
+import { Translator } from '../../../utils/i18n';
+import { ContributionDonationEntry } from '../contribution/contribution.types';
+import { ContributorDonationCertificate } from '../contributor/contributor.types';
+
+const ASSET_DIR = path.join(process.cwd(), 'public', 'assets');
+
+export class DonationCertificateWriter {
```
rawi96, line 12: pls rename the file from DonationCertificateWriter to donation-certificate-writer

website/infra/variables.tf
```diff
@@ -34,8 +34,8 @@ variable "website_domain" {
   type        = string
 }
 
-variable "exchange_rate_import_api_key" {
+variable "scheduler_api_key" {
   description = "API key for the exchange rate import API endpoint"
```
rawi96: pls also update the description

shared/src/database/services/contributor/contributor.service.ts
fabricem90: Hmm the function is a bit more specific because it either returns contributors by IDs or (if IDs are not provided) returns all swiss contributors for donation certificate generation

ssandino: We can remove the Swiss specificity since all contributors should be able to download their donation certificate. 

And we can get rid of the address requirement in the text, so it passes for all.

fabricem90: @ssandino do you mean by `address requirement in the text` the CH country assertion/check or the rendering of the address in the PDF?

fabricem90: If we want to generate certificates also for non-swiss residents, we'd maybe also have to adjust the PDF content since currently we have switzerland specific texts (eg. `Please refer to the cantonal tax regulations for more information on how to deduct your
donations.`)

shared/src/database/services/contribution/contribution.service.ts
```diff
@@ -162,6 +163,47 @@ export class ContributionService extends BaseService {
 		}
 	}
 
+	async getForContributorsAndYear(
+		contributorsIds: string[],
+		year: number,
+	): Promise<ServiceResult<ContributionDonationEntry[]>> {
+		try {
+			const result = await this.db.contribution.findMany({
```
rawi96, line 175: I think we already have date-fns as dependency. So we could smth like this


```
import { startOfYear, endOfYear } from 'date-fns';

const start = startOfYear(new Date(year, 0, 1));
const end = endOfYear(new Date(year, 0, 1));

const result = await this.db.contribution.findMany({
	where: {
		contributorId: { in: contributorsIds },
		createdAt: {
			gte: start,
			lte: end,
		},
	},
});

```

shared/src/database/services/contributor/contributor.service.ts
```diff
@@ -199,6 +200,47 @@ export class ContributorService extends BaseService {
...
 	}
 
+	async getByIds(contributorIds?: string[]): Promise<ServiceResult<ContributorDonationCertificate[]>> {
+		try {
+			const result = await this.db.contributor.findMany({
+				where: contributorIds
+					? {
+							id: { in: contributorIds },
+						}
+					: undefined,
```
rawi96: This is confusing, could we do this instead?

```
where: contributorIds && contributorIds.length > 0
	? { id: { in: contributorIds } }
	: {}
```

Should be the same but more explizit

shared/src/database/services/donation-certificate/donation-certificate-writer.ts
```diff
@@ -0,0 +1,161 @@
...
+	public readonly donator: ContributorDonationCertificate;
+	public readonly contributions: ContributionDonationEntry[];
+
+	constructor(donator: ContributorDonationCertificate, contributions: ContributionDonationEntry[], year: number) {
+		this.donator = donator;
+		this.year = year;
+		this.contributions = contributions;
+	}
+
+	private getContributionsByCurrency = async (contributions: ContributionDonationEntry[], year: number) => {
```
rawi96: could we rename this to `groupContributionsByCurrency`?

PR #1469, opened 2025-12-01
feat: migrate surveys

fabricem90:


website/src/app/[lang]/[region]/survey/[recipient]/[survey]/page.tsx
Comment (rawi96) [inline website/src/app/[lang]/[region]/survey/[recipient]/[survey]/page.tsx]: Could we simplify this a little?
I think we dont need states for email and password. and also the useEffect could be removed?

```
const searchParams = useSearchParams()
const email = searchParams.get('email')
const password = searchParams.get('password')

```
Should be the same or not?

website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts
```diff
@@ -0,0 +1,82 @@
...
+			const idToken = await userCredential.user.getIdToken(true);
+
+			const result = await createSessionAction(idToken);
+			if (!result.success) {
+				setHasError(true);
+				return false;
+			}
+			return result.success;
+		} catch (error) {
+			console.log(error);
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts]: logger.error

website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts
```diff
@@ -0,0 +1,82 @@
...
+
+	const saveSurvey = (surveyId: string, survey: Model, status: SurveyStatus, retryCount = 0) => {
+		const data = survey.data;
+		data.pageNo = survey.currentPageNo;
+		saveChanges(surveyId, {
+			data: data,
+			status: status,
+			completedAt: status == SurveyStatus.completed ? new Date(Date.now()) : null,
+		})
+			.then(() => {
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts]: could we use async await syntax?

website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts
```diff
@@ -0,0 +1,82 @@
...
+				console.log('saved successfully');
+			})
+			.catch(() => {
+				if (retryCount >= 2) {
+					setHasError(true);
+					logout();
+					console.log('error saving, abording');
+					return;
+				}
+				console.log('error saving, retrying');
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/survey/[recipient]/[survey]/use-survey.ts]: remove console logs or use the logger

website/src/lib/server-actions/survey-actions.ts
```diff
@@ -58,3 +59,25 @@ export async function generateSurveysAction() {
 	revalidatePath('/portal/management/surveys');
 	return result;
 }
+
+export async function getByIdAndRecipient(surveyId: string, recipientId: string) {
+	const survey = await getCurrentSurvey();
+	if (!survey || survey.id !== surveyId || survey.recipientId !== recipientId) {
+		throw new Error('Unauthorized');
```
rawi96, line 66: we mix now the result object of the service with throwing errors

so this server actions here could either directly throw errors, or return a result which could potentially contain success false.

I would align on one pattern.
either always throw or always return success: true/false

website/src/lib/server-actions/survey-actions.ts
```diff
@@ -58,3 +59,25 @@ export async function generateSurveysAction() {
...
+	const service = new SurveyService();
+
+	const result = await service.getByIdAndRecipient(surveyId, recipientId);
+	return result;
+}
+
+export async function saveChanges(surveyId: string, input: SurveyUpdateInput) {
+	const survey = await getCurrentSurvey();
+	if (!survey || survey.id !== surveyId) {
+		throw new Error('Unauthorized');
```
rawi96, line 80: same here

website/src/lib/services/survey/survey.service.ts
```diff
@@ -426,4 +427,97 @@ export class SurveyService extends BaseService {
 			return this.resultFail('Could not fetch surveys');
 		}
 	}
+
+	async getByEmail(email: string): Promise<ServiceResult<SurveyPayload>> {
+		try {
+			const surveys = await this.db.survey.findFirst({
```
rawi96: is findFirst correct?
should we maybe make the accessEmail unique and then use findUnique?

website/src/lib/services/survey/survey.service.ts
```diff
@@ -426,4 +427,97 @@ export class SurveyService extends BaseService {
 			return this.resultFail('Could not fetch surveys');
 		}
 	}
+
+	async getByEmail(email: string): Promise<ServiceResult<SurveyPayload>> {
```
rawi96: could we rename to getByAccessEmail to make more clear its this "pseudo" email and not the real email of the recipient

website/src/app/[lang]/[region]/survey/[recipient]/[survey]/page.tsx
Comment (fabricem90) [reply website/src/app/[lang]/[region]/survey/[recipient]/[survey]/page.tsx]: True. Maybe the idea was, that search params could be reactive when/if changed through the application during runtime. But we wont have this case for the survey

PR #1523, opened 2025-12-05
feat: creatable contributors

rawi96:


website/src/app/portal/management/contributors/contributors-form.tsx
```diff
@@ -99,8 +113,10 @@ export default function ContributorsForm({
 
 	useEffect(() => {
 		if (contributorId) {
-			// Load contributor
-			startTransition(async () => await loadContributor(contributorId));
```
fabricem90, line 103: without `startTransition`, loading state will be lost

PR #1567, opened 2025-12-16
feat: dashboard subscription improvements

fabricem90:
## Summary by CodeRabbit

* **New Features**
  * Visual subscription status badge (active, canceled, paused)
  * Payment method shown in subscriptions list

* **Improvements**
  * Subscriptions table shows creation date and payment method
  * Contribution amounts use enhanced currency formatting
  * Locale strings updated for subscription labels

* **Bug Fixes**
  * Post-donation redirect now goes to the appropriate dashboard section (subscriptions or contributions)

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/website/src/app/[lang]/[region]/dashboard/profile/page.tsx
+++ b/website/src/app/[lang]/[region]/dashboard/profile/page.tsx
@@ -2,6 +2,7 @@ import { Card } from '@/components/card';
 import { getAuthenticatedContributorOrRedirect } from '@/lib/firebase/current-contributor';
 import { Translator } from '@/lib/i18n/translator';
 import { WebsiteLanguage } from '@/lib/i18n/utils';
+import { getActiveSubscription } from '@/lib/server-actions/newsletter-actions';
 import { COUNTRY_CODES, CountryCode } from '@/lib/types/country';
 import { DefaultPageProps } from '../..';
 import { ProfileForm, ProfileFormTranslations } from './profile-form';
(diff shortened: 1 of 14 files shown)


website/src/app/[lang]/[region]/(blue-theme)/donate/success/stripe/[session]/page.tsx
```diff
@@ -34,7 +34,7 @@ export default async function Page({ params }: StripeSuccessPageProps) {
 	}
 
 	if (contributorResult.data && !contributorResult.data.needsOnboarding) {
-		redirect(`/${lang}/${region}/dashboard/contributions`);
+		redirect(`/${lang}/${region}/dashboard/subscriptions`);
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/(blue-theme)/donate/success/stripe/[session]/page.tsx]: i think the idea is to check if a contributor did a one time donation and if yes redirect to contributions and if its e.g. a monthly subscription redirect to subscriptions. 

I don't know if its even possible?

website/src/lib/services/stripe/stripe.service.ts
```diff
@@ -289,6 +294,21 @@ export class StripeService extends BaseService {
...
+		const titleCase = (s: string) =>
+			s.replace(/^_*(.)|_+(.)/g, (s, c, d) => (c ? c.toUpperCase() : ' ' + d.toUpperCase()));
+		if (paymentMethod.type === 'card' && paymentMethod.card) {
+			return {
+				type: 'card',
+				label: titleCase(paymentMethod.card.brand),
+			};
+		}
+		return {
+			type: 'other',
```
rawi96, line 307: Is there even another type then card in stripe?
which one? twint? Could we then not render twint instead of other?

fabricem90: On Safari/iOS there's Apple Pay, Twint is only available for one-time payments in Switzerland. Potentially there is also Paypal and dozens of other Payment Providers depending on country/region which could be activated. Therefore I thought its the easiest to just group them in "others" for now. 

Name of payment method such as "Twint" will still be shown, this type is currently only used for Icon rendering

website/src/app/[lang]/[region]/(blue-theme)/donate/success/stripe/[session]/page.tsx
Comment (fabricem90) [reply website/src/app/[lang]/[region]/(blue-theme)/donate/success/stripe/[session]/page.tsx]: yes we can :)

PR #1684, opened 2026-01-31
Unify country codes (only use ISO 3166-1 alpha-2 codes)

mkue:
When I started building a new transparency page, I noticed that a new ISO code for countries was introduced (`iso-countries.ts` file), which messed up the CountryFlag component when using the country codes in the address (2-character country code). I see no reason to use 2-character and 3-character codes, so I got rid of the 3-character codes again because the addresses in the db use and the translation files already use two digit codes.

This PR also adds a super basic initial transparency page on the new website and a CLAUDE.md file.

The transparency page code is just POC and will probably completely change again.

## Summary by CodeRabbit

* **New Features**
  * Added a Transparency → Finances page with totals, monthly contributions chart, and a "Top Contributing Countries" view (localized currency, flags, progress bars).

* **Documentation**
  * Added comprehensive project-wide guidelines and architecture documentation.

* **Chores**
  * Standardized ISO-2 country code handling across data, types and DB schema; updated DB tooling scripts and seed data.

* **UX**
  * Country selector updated to use curated options/combobox and locale-aware country name lookups.

* **Tests**
  * Updated e2e expectation to use ISO-2 country codes.

### DIFF
--- a/website/src/app/portal/admin/candidates/candidate-form-helpers.ts
+++ b/website/src/app/portal/admin/candidates/candidate-form-helpers.ts
@@ -155,13 +155,15 @@ export function buildUpdateCandidateInput(
 				data: {
 					...buildCommonContactData(contactFields),
 					phone: contactPhoneWriteOperation,
-					address: {
-						upsert: {
-							update: addressUpdateOperation,
-							create: addressUpdateOperation,
-							where: { id: candidate.contact.address?.id },
+					...(addressUpdateOperation && {
+						address: {
+							upsert: {
+								update: addressUpdateOperation,
+								create: addressUpdateOperation,
+								where: { id: candidate.contact.address?.id },
+							},
 						},
-					},
+					}),
 				},
 				where: { id: candidate.contact.id },
 			},
@@ -174,6 +176,7 @@ export function buildCreateCandidateInput(
 	contactFields: { [key: string]: FormField },
 ): CandidateCreateInput {
 	const paymentInfoFields = schema.fields.paymentInformation.fields;
+	const addressInput = buildAddressInput(contactFields);
 
 	return {
 		status: schema.fields.status.value,
@@ -191,7 +194,9 @@ export function buildCreateCandidateInput(
 			create: {
 				...buildCommonContactData(contactFields),
 				phone: contactFields.phone.value ? { create: { number: contactFields.phone.value } } : undefined,
-				address: { create: buildAddressInput(contactFields) },
+				...(addressInput && {
+					address: { create: addressInput },
+				}),
 			},
 		},
 	};
--- a/website/src/app/portal/admin/local-partners/local-partners-form-helper.ts
+++ b/website/src/app/portal/admin/local-partners/local-partners-form-helper.ts
@@ -21,8 +21,10 @@ export function buildUpdateLocalPartnerInput(
 				},
 			}
 		: undefined;
+
 	// Contact Address Upsert Logic
 	const addressUpdate = buildAddressInput(contactFields);
+
 	return {
 		id: localPartner.id,
 		name: schema.fields.name.value,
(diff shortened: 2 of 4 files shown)


website/src/lib/database/enums.prisma
```diff
@@ -0,0 +1,252 @@
+/// ISO 3166-1 alpha-2 country codes
+enum CountryCode {
```
rawi96, line 127: All other enums are defined in schema.prisma.
We should either move this enum there as well, or extract all enums into separate enum files for consistency.

website/src/app/[lang]/[region]/new-website/transparency/finances/[currency]/page.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { Currency } from '@/lib/types/currency';
+import { DateTime } from 'luxon';
+import { CountriesSection } from './(sections)/countries-section';
+import { TimeSeriesSection } from './(sections)/time-series-section';
+import { TotalsSection } from './(sections)/totals-section';
+
+export const revalidate = 3600;
+export const generateStaticParams = () => websiteCurrencies.map((currency) => ({ currency: currency.toLowerCase() }));
+
+interface TransparencyFinancesParams extends DefaultParams {
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/new-website/transparency/finances/[currency]/page.tsx]: Please replace interface with a type. Because ee always use type because it’s more flexible than interface and avoids accidental declaration merging.


```
type TransparencyFinancesParams = DefaultParams & {
	currency: Currency;
};
```

website/src/components/profile-form/submit.ts
```diff
@@ -32,24 +32,26 @@ export async function submitProfileForm(
...
-								create: {
-									street: values.street ?? '',
-									number: values.number ?? '',
-									city: values.city ?? '',
-									zip: values.zip ?? '',
-									country: values.country ?? '',
-								},
-							},
-						},
+						address: values.country
```
rawi96: Why is the country needed to update street, nr, city etc? 🤔

website/src/components/profile-form/submit.ts
```diff
@@ -70,24 +72,26 @@ export async function submitProfileForm(
...
```
rawi96: same here

website/src/components/profile-form/submit.ts
```diff
@@ -102,12 +106,14 @@ export async function submitProfileForm(
...
 		language: values.language,
 		organizationId: values.organizationId,
-		address: {
-			street: values.street ?? '',
-			number: values.number ?? '',
-			city: values.city ?? '',
-			zip: values.zip ?? '',
-			country: values.country ?? '',
-		},
+		address: values.country
```
rawi96: same here

website/src/lib/database/enums.prisma
rawi96: The advantage of having everything in one file is that you can use tools like [visualizers](https://prismaliser.app/) more easily.

But I'm also fine if we start to split all the enums into files.

website/src/lib/services/transparency/transparency.service.ts
```diff
@@ -0,0 +1,94 @@
...
+import { ServiceResult } from '../core/base.types';
+import {
+	ContributionsByCountry,
+	ContributionTimeRange,
+	TimeRange,
+	TransparencyData,
+	TransparencyTotals,
+} from './transparency.types';
+
+export class TransparencyService extends BaseService {
```
rawi96, line 13: We should avoid $queryRaw here and use Prisma’s query API instead—so far we don’t use raw queries anywhere else in the project, and they should generally be reserved for exceptional cases only. Prisma queries are type-safe and refactor-friendly; for example, a renamed DB column would be caught at compile time instead of failing at runtime.

website/package.json
```diff
@@ -121,6 +121,7 @@
 		"typescript": "^5.6.2"
 	},
 	"prisma": {
+		"schema": "./src/lib/database",
```
rawi96, line 125: ❤️

website/src/lib/database/enums.prisma
mkue: the enum is just huge and imo the file becomes harder and harder to read.

website/src/components/profile-form/submit.ts
mkue: The issue is that the country can't be empty anymore with the db enum. I updated the form now a little so it makes more sense now.

website/src/lib/services/transparency/transparency.service.ts
mkue: I changed it and generally agree. For some statistics, it might make sense to make exceptions in favor of performance and readability. The screens should then definitely be covered by (e.g. screenshot) tests to make sure things don't break in production.

PR #1892, opened 2026-03-15
Mobile App: Use new API (without offline support)

KarinBerg:
- This PR includes all changes using the new mobile API https://staging.socialincome.org/v1/api-docs
- The app is now considering the actual program length and payment interval from the model "Program".
- Remove not needed anymore SLL currency conversion edge case in past payments calculation.
- Handle login case "no recipient found": Show user message and go back to Welcome Page if there is no recipient found.
- Updated to Flutter 3.41.4
- Updated Flutter packages
- Moved from json_serializable and equatable to dart_mappable
- Add Flutter Version Manager (FVM) support
- It does NOT contain the changes to support offline usage (see https://github.com/socialincome-san/public/pull/1894 for this)


recipients_app/pubspec.yaml
```diff
@@ -2,7 +2,7 @@ name: app
 description: Social Income App
 
 publish_to: "none"
-version: 1.1.20+1 # The build number is set in Codemagic automatically.
+version: 1.2.0+1 # The build number is set in Codemagic automatically.
```
KarinBerg: I suggest to change the version number to 2.0.0 because this is a non-backward compatible change which uses the new API and once we deactivate Firestore completely, users will need to use the new app version.

recipients_app/pubspec.yaml
```diff
@@ -75,15 +66,14 @@ flutter:
           weight: 500
         - asset: assets/fonts/Unica77LL-Bold.otf
           weight: 700
+# flutter_native_splash:
```
KarinBerg: @novas1r1 Why did you comment out flutter_native_splash. Do we not need it anymore. Why?

recipients_app/lib/core/cubits/payment/payouts_cubit.dart
```diff
@@ -0,0 +1,283 @@
...
+import "package:dart_mappable/dart_mappable.dart";
+import "package:firebase_core/firebase_core.dart";
+import "package:flutter/material.dart";
+import "package:flutter_bloc/flutter_bloc.dart";
+
+part "payouts_cubit.mapper.dart";
+part "payouts_state.dart";
+
+const int kMaxReviewDays = 10;
+//TODO(migration): get real amount from backend (Model Programm) and remove this constant
```
novas1r1: @KarinBerg this is tackled in the next pr?

KarinBerg: Not yet, this is the last open task.

recipients_app/lib/core/cubits/payment/payouts_cubit.dart
```diff
@@ -0,0 +1,283 @@
...
+    emit(state.copyWith(status: PayoutsStatus.loading));
+
+    try {
+      final payouts = await paymentRepository.fetchPayouts();
+
+      emit(
+        state.copyWith(
+          status: PayoutsStatus.success,
+          payoutsUiState: await _mapPayoutsUiState(payouts),
+        ),
```
novas1r1, line 46: @KarinBerg maybe we could set ex to null here. Could also be done in loading state

KarinBerg: To make it explicit? Or for what reason.

recipients_app/lib/core/cubits/payment/payouts_cubit.dart
KarinBerg: Last commit "Use new payment amount and program duration from Model "Program" in the app" contains now these required changes.

recipients_app/pubspec.yaml
KarinBerg: For now, I added it again

recipients_app/lib/core/cubits/dashboard_card_manager/dashboard_card_manager_cubit.dart
```diff
@@ -1,64 +1,45 @@
...
+      // That means the payment number is always the same as the firebase auth phone number and so it can not be null.
+      final contactPhoneNumber = recipient.contact.phone?.number;
+      final paymentPhoneNumber = recipient.paymentInformation?.phone.number;
 
       if (contactPhoneNumber == null && paymentPhoneNumber != null) {
         final contactPhoneCard = DashboardCard(
           title: "My Profile",
-          message:
-              "Is your payment phone number (${paymentPhoneNumber.phoneNumber}) also your contact phone number?",
+          message: "Is your payment phone number ($paymentPhoneNumber) also your contact phone number?",
```
MDikkii, line 42: Probably not for this PR but it should be added to i10n.

recipients_app/lib/core/cubits/dashboard_card_manager/dashboard_card_manager_state.dart
```diff
@@ -1,8 +1,16 @@
 part of "dashboard_card_manager_cubit.dart";
 
-enum DashboardCardManagerStatus { initial, loading, loaded, error }
+enum DashboardCardManagerStatus {
```
MDikkii, line 3: Is it ok without `@MappableEnum`? Some of the enums need it, and some don't? Is it because of the local logic for cards?

MDikkii (approved):

recipients_app/lib/core/cubits/dashboard_card_manager/dashboard_card_manager_cubit.dart
KarinBerg: Ups :)

recipients_app/lib/core/cubits/dashboard_card_manager/dashboard_card_manager_state.dart
KarinBerg: Good point, will check that before merge.

PR #2158, opened 2026-06-18
feat: add dedicated login page

### LINKED ISSUE #2150: create a dedicated login page
the new website has a login flyout in the navigation which works nicely.

But to eg redirect a user to a dedicated login page after logout etc it would be nice to have a login page.

design could be smth from shadcn. Maybe a few parts of the login flyout could be reused.

https://shadcnstudio.com/blocks/marketing-ui/login-page

nschaer92:
## Summary by CodeRabbit

* **New Features**
  * Added a dedicated magic-link login form with localized email input, “sent” confirmation, retry, and support contact.
* **Refactor**
  * Replaced the login flyout’s inline magic-link flow with the shared magic-link login form.
* **Navigation / Login Behavior**
  * Improved non-email-link login routing: when sessions exist, users are redirected to the computed post-login destination; otherwise the localized login screen is shown.
* **Localization**
  * Removed the “get-started” login message from English, German, French, and Italian locales.

### DIFF
--- a/website/src/app/[lang]/[region]/login/login-page-content.tsx
+++ b/website/src/app/[lang]/[region]/login/login-page-content.tsx
@@ -1,0 +1,24 @@
+'use client';
+
+import { MagicLinkLoginForm } from '@/components/login/magic-link-login-form';
+import { useTranslator } from '@/lib/hooks/useTranslator';
+import { WebsiteLanguage } from '@/lib/i18n/utils';
+
+type Props = {
+	lang: WebsiteLanguage;
+};
+
+export const LoginPageContent = ({ lang }: Props) => {
+	const translator = useTranslator(lang, 'website-login');
+
+	return (
+		<div className="flex min-h-[40vh] flex-1 flex-col items-center justify-center px-4">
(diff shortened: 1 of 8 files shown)


website/src/app/[lang]/[region]/login/login-page-content.tsx
```diff
@@ -0,0 +1,22 @@
+'use client';
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/login/login-page-content.tsx]: why client-side?

website/src/app/[lang]/[region]/login/login-page-content.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { MagicLinkLoginForm } from '@/components/login/magic-link-login-form';
+import { useTranslator } from '@/lib/hooks/useTranslator';
+import { WebsiteLanguage } from '@/lib/i18n/utils';
+
+type Props = {
+	lang: WebsiteLanguage;
+};
+
+export const LoginPageContent = ({ lang }: Props) => {
+	const translator = useTranslator(lang, 'website-login');
```
Comment (rawi96) [inline website/src/app/[lang]/[region]/login/login-page-content.tsx]: use the server side translator:
const translator = await Translator.getInstance({ language: lang as WebsiteLanguage, namespaces: ['website-login'] });

rawi96 (commented):
i think <loginPageContent should be a server component.

website/src/app/[lang]/[region]/login/login-page-content.tsx
Comment (nschaer92) [reply website/src/app/[lang]/[region]/login/login-page-content.tsx]: good catch. implemented as server-component

PR #2267, opened 2026-06-24
feat: add options to disable spacing and add new spacer block

rawi96:
## Summary by CodeRabbit

* **New Features**
  * Added a new spacer block for creating adjustable vertical spacing on pages.
  * Expanded block spacing controls so content blocks can independently remove top or bottom margins.
  * Updated page rendering to support the new spacer block type.

* **Bug Fixes**
  * Improved image-text block rendering by using more reliable content handling.
  * Preserved existing block behavior while making spacing options more flexible.


website/src/generated/storyblok/types/109655/storyblok-components.d.ts
```diff
@@ -449,6 +452,13 @@ export interface ReferencesGroup {
   [k: string]: unknown;
 }
 
+export interface Spacer {
+  size?: "" | "xs" | "s" | "m" | "l" | "xl";
```
MumenthalerManuel: make size required and remove default size in code -> cms can provide default instead

PR #2409, opened 2026-07-18
Journal: fixing quoted text

ssandino:


ssandino:
Before: 
Screenshot 2026-07-18 at 16 06 15

After: 
Screenshot 2026-07-18 at 16 29 02

PR #2478, opened 2026-07-23
Fix focus filter widths on small screens

### LINKED ISSUE #2477: Make filter dropdowns fill the available width
When two filter dropdowns are displayed on the same row on a small screen (like [here](https://socialincome.org/en/ch/focuses)), they should expand to use the full available width instead of keeping a fixed width and leaving empty space on the right.

This creates a cleaner, more balanced layout and allows longer filter labels to fit more easily.

**Expected behavior**

* If two dropdowns are shown in one row on small screen, they should together occupy 100% of the available width.
* Both dropdowns should flex automatically so that the longer label gets more space while the second dropdown fills the remaining width. And of course have a small gap in between.
* There should be no unused whitespace at the end of the row.

Image

sean-camara:
## Summary

- Make the two Focuses filter dropdowns fill the available row width on small screens
- Preserve the gap between the controls
- Keep the existing larger-screen layout behavior

Closes #2477

## Testing

- Reproduced the current layout at 320px, 375px, and 412px using the public Focuses page
- Verified the touched files with Prettier and ESLint
- Ran 
pm run typecheck and 
pm run build; both are blocked by an incomplete date-fns package in the local installation (missing locale files), unrelated to this change
- Full workspace formatter and lint checks report pre-existing repository-wide issues unrelated to the touched files

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the filter/search control row layout to stay on a single line without wrapping, keeping controls aligned.
  * Updated filter dropdowns to better utilize available space and prevent long selected labels from overflowing, with improved truncation and icon sizing.


ssandino:
@sean-camara there seems to be a conflict (see above)

MumenthalerManuel:
@sean-camara I just updated the filter code and moved it to a shared filter-dropdown file. Can you replicate the change here in the new filter-dropdown? This applies the updated design to all pages that use a filter dropdown. Change is looking good otherwise, I'll approve as soon as the conflict is resolved

sean-camara:
Thanks! I've merged the latest main, resolved the conflict, and moved the responsive sizing and label truncation changes into the shared FilterDropdown so they apply to all pages using it. The branch is mergeable now, and the updated checks are running.

MumenthalerManuel (approved):

PR #2480, opened 2026-07-23
Align local partner social links

### LINKED ISSUE #2479: Content alignment local program partner details
The social media links (LinkedIn, Instagram, Facebook, YouTube, etc.) of a [local program partner ](https://socialincome.org/en/ch/local-partners/aurora-foundation)are currently displayed as another row within the information grid, making the layout feel unbalanced.

A cleaner layout would be to: 

* the pills with the links should go on the same row as the label "Online" (like we handle other information: Location (left), City (right) is also on the same row, see printscreen below)
* write "Learn more" instead of "Online" (clearer wording)
* add a small arrow after word in pill to show that it's an external link, so instead of `(Facebook)` it should be `(Facebook ↗︎)` 

**Proposed layout**

* Lay out the pills responsively:
    * If there is enough horizontal space, display all pills on a single row. (same row as label left)
    * Otherwise, wrap them onto multiple rows as needed (e.g. on mobile).

**Acceptance criteria**

* Label text change and ↗︎ icon in pill
* Social media pills are moved on same row as label
* Pills wrap naturally based on the available width
* The layout remains responsive across desktop, tablet, and mobile

Printscreen of issue: 
Image

sean-camara:
## Summary

- Place local-partner social-link pills beside the Learn more label
- Allow pills to wrap naturally at narrower widths
- Add an external-link indicator to each social-link pill

Closes #2479

## Testing

- npm run typecheck
- Prettier and ESLint on the changed component
- git diff --check

## Summary by CodeRabbit

* **Style**
  * Updated local partner information cards with a more compact “Learn more” link layout.
  * Added a trailing arrow icon to external links to better indicate they open in a new tab.


MumenthalerManuel:
Hi @sean-camara Thanks for the PR. I'm investigating the problem with the e2e test that is failing. This does not happen because of your change but because the workflow does not have access to secrets when run from a fork and therefore it does not have a storyblok access token. I'll merge the PR asap and we'll look for a way to handle this problem next week.

sean-camara:
Thanks for investigating and confirming! That makes sense. I appreciate the update.

PR #2489, opened 2026-07-23
Mobile App: Remove dashboard item which asks the user if he want to set the payment number as contact number

KarinBerg:
We removed this because in the portal we use the payment number as contact number if no explicit contact number is set by the user. And the payment number must always be there.

## Summary by CodeRabbit

- **Improvements**
  - Contact phone validation now allows empty input and shows a clearer numeric-only message when non-empty text isn’t a valid number.
  - Contact phone is updated only on valid input, and only when it differs from the saved payment phone (handles leading “+” consistently).
  - Dashboard now renders header items plus surveys; the prior contact-phone dashboard card flow was removed, with an empty state when no surveys are available.
- **Documentation**
  - Updated the Profile/Account manual test plan to reflect the revised saving and offline-check rules.
- **Localization**
  - Refreshed the contact phone error text in English and Krio. 
- **Chores**
  - Updated demo recipient contact phone to be blank.


ssandino:
> We removed this because in the portal we use the payment number as contact number if no explicit contact number is set by the user. And the payment number must always be there.

This makes totally sense!

PR #2527, opened 2026-07-26
Web: Person grid component

ssandino:
## Summary by CodeRabbit

* **New Features**
  * Added a new **person-grid** content block with responsive person cards, interactive search/sort/filter UI, and localized empty-state messaging.
  * Added **primary role-label** mapping across person cards, carousels, and journal pages.
  * Added **“volunteering since”** badges with localized duration formatting and hover details.
* **Bug Fixes**
  * Improved country-office person results and ensured role labels display correctly in carousels.
  * Fixed **Spacer** height sizing when the configured size is empty.
* **Localization**
  * Added **person-grid** UI translations (en, de, fr, it).


ssandino:
Added all different options and preferences so the site [/team](https://socialincome.org/en/ch/team) can be rebuild with the actual design elements and storyblok (see localhost:3000/en/int/people): 

Screenshot 2026-07-27 at 00 31 58
Screenshot 2026-07-27 at 00 32 14

website/src/components/content-blocks/person-grid.tsx
```diff
@@ -0,0 +1,168 @@
...
+	region: WebsiteRegion;
+};
+
+const matchesStatusFilter = (person: ISbStoryData<Person>, statusFilter: string) =>
+	statusFilter === 'all' || statusFilter === person.content.volunteerStatus;
+
+const matchesRoleFilter = (person: ISbStoryData<Person>, roleFilterCodes: string[]) =>
+	roleFilterCodes.length === 0 || personHasRole(person, roleFilterCodes);
+
+// Excluded roles are enforced ahead of every other branch (manual picks, country office, role/status
```
MumenthalerManuel: Those are very hard to read. Could you change it to something that does not contain the double negation in the function? Something like this would be a lot easier to read. 

const isRoleExcluded = (person: ISbStoryData<Person>, roleExcludeCodes: string[]) =>
  roleExcludeCodes.length > 0 && personHasRole(person, roleExcludeCodes);

const isCountryOfficeMember = (person: ISbStoryData<Person>) =>
  Boolean(person.content.countryOffice?.length);

// usage
persons.filter(
  (person) =>
    !isRoleExcluded(person, roleExcludeCodes) &&
    !(excludeCountryOfficeMembers && isCountryOfficeMember(person)),
);

website/src/components/content-blocks/person-grid.tsx
```diff
@@ -0,0 +1,168 @@
...
+	const roleExcludeCodes = toStringArray(blok.roleExcludeFilter);
+	const excludeCountryOfficeMembers = blok.excludeCountryOfficeMembers ?? false;
+
+	const statusFilter = blok.statusFilter ?? 'all';
+	const limit = Number(blok.limit) || 0;
+	const centerLastRow = blok.centerLastRow ?? false;
+	const cardSize = blok.cardSize === 'small' || blok.cardSize === 'medium' ? blok.cardSize : 'large';
+	const cardLink = blok.cardLink === 'none' || blok.cardLink === 'dialog' ? blok.cardLink : 'personPage';
+	const showVolunteerDuration = blok.showVolunteerDuration ?? false;
+	const showSearch = blok.showSearch ?? false;
```
MumenthalerManuel, line 43: Does this need to be 3 properties or could it just be one toggle for the block?

website/src/components/content-blocks/person-grid.tsx
```diff
@@ -0,0 +1,168 @@
...
+	const allPersons = (personsResult.success ? personsResult.data : []).filter(
+		(person) =>
+			matchesRoleExcludeFilter(person, roleExcludeCodes) &&
+			matchesCountryOfficeExcludeFilter(person, excludeCountryOfficeMembers),
+	);
+
+	const persons = manualUuids.length
+		? allPersons
+		: isInteractive
+			? allPersons.filter((person) => matchesRoleFilter(person, roleFilterCodes))
```
MumenthalerManuel: Is there a reason status filter is not applied if "isInteractive" is true here? With this implementation you would expect the statuFilter to do filter out "inactive" users if the filter is set but they would still show up in the list if the search bar is enabled for example.

website/src/components/content-blocks/person-grid.tsx
```diff
@@ -0,0 +1,168 @@
...
+			: allPersons.filter(
+					(person) => matchesRoleFilter(person, roleFilterCodes) && matchesStatusFilter(person, statusFilter),
+				);
+	const limitedPersons = !isInteractive && limit > 0 ? persons.slice(0, limit) : persons;
```
MumenthalerManuel: If you add a "max count" style property I would always apply it to the displayed grid and not just in some cases. Maybe we could drop this property? There are not too many voulenteers and you can control the displayed content with the filters to make sure there is no accidental "show 50 people" case. Slicing them off without a "show more" button could also be weird since you don't know which voulenteers would be shown and which would be cut off.

website/src/components/storyblok/shared/person-card-grid.tsx
```diff
@@ -0,0 +1,82 @@
...
+	const { personCardSize, gridCols, flexItemWidth } = CARD_SIZE_CONFIG[cardSize];
+	// 'dialog' isn't implemented yet — falls back to no link until that's built.
+	const getHref = (person: ISbStoryData<Person>) =>
+		cardLink === 'personPage' ? createWebsitePersonLink(person.slug, lang, region) : undefined;
+	const volunteerDuration = volunteerDurationTranslations
+		? { lang, translations: volunteerDurationTranslations }
+		: undefined;
+
+	return (
+		<ul className={centerLastRow ? 'flex flex-wrap justify-center gap-6' : cn('grid gap-6', gridCols)}>
```
MumenthalerManuel: I would rather decide between using flex or using grid here and not switch it based on a boolean. This way the layout just needs to support both - all changes will also have to make sure that both are working. 

If the centered last column is important I would drop the grid and always use the flex with. If it is not that important I would keep the grid and remove the flexWidth part.

website/src/components/storyblok/shared/person-card-grid.tsx
```diff
@@ -0,0 +1,82 @@
...
+	personCardSize: 'default' | 'small' | 'compact';
+	gridCols: string;
+	flexItemWidth: string;
+};
+
+const CARD_SIZE_CONFIG: Record<CardSize, CardSizeConfig> = {
+	large: {
+		personCardSize: 'default',
+		gridCols: 'grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 xl:grid-cols-4',
+		flexItemWidth: 'w-full sm:w-[calc(50%-0.75rem)] lg:w-[calc(33.333%-1rem)] xl:w-[calc(25%-1.125rem)]',
```
MumenthalerManuel: If you keep the flexItemWidth, could you use something like "basis-1/2 lg:basis-1/4 xl:basis-1/5 shrink-0" instead of the calc(50%...) calculations?

website/src/components/storyblok/shared/person-card-grid.tsx
```diff
@@ -0,0 +1,82 @@
...
+	centerLastRow?: boolean;
+	cardSize?: CardSize;
+	cardLink?: CardLink;
+	// Presence enables the "volunteering since" pill on the cards.
+	volunteerDurationTranslations?: VolunteerDurationTranslations;
+	roleLabels?: Record<string, string>;
+};
+
+type CardSizeConfig = {
+	personCardSize: 'default' | 'small' | 'compact';
```
MumenthalerManuel: Why are the CardSizes renamed here? Could we just pass through the large / medium / small labels instead?

website/src/components/storyblok/shared/person-card.tsx
```diff
@@ -8,28 +16,97 @@ import NextLink from 'next/link';
...
+			? getDurationLabels(volunteerSince, volunteerDuration)
+			: null;
 
 	const card = (
 		<div
 			className={cn(
 				'bg-card w-full overflow-hidden rounded-xl shadow-[0px_4px_28px_0px_rgba(0,30,101,0.07)]',
 				isSmall ? 'max-w-[260px] p-2.5' : 'max-w-[305px] p-3',
-				href && 'transition-transform hover:scale-[1.01]',
+				href && 'transform-gpu transition-transform hover:scale-[1.01]',
```
MumenthalerManuel: Why is transform-gpu necessary here? For the hover scale transform this should not be needed.

website/src/components/storyblok/shared/person-card.tsx
```diff
@@ -39,6 +116,18 @@ export const PersonCard = ({ person, href, size = 'default', className }: Props)
...
+				{duration ? (
+					<Badge
+						variant="default"
+						// Hover-only content is invisible to assistive tech, so the date rides along as the accessible
+						// description; an aria-label would instead replace the duration as the accessible name.
+						title={duration.since}
+						className="group/duration text-foreground absolute top-3 left-3 z-20 border-white/40 bg-white/80 whitespace-nowrap backdrop-blur-sm"
+					>
+						<span className="group-hover/duration:hidden">{duration.label}</span>
+						<span className="hidden group-hover/duration:inline">{duration.since}</span>
```
MumenthalerManuel, line 128: This will never be visible on mobile phones because they don't use hover.

website/src/components/storyblok/shared/person-grid-interactive.tsx
```diff
@@ -0,0 +1,437 @@
...
+	showFilterPills,
+	defaultStatus,
+	limit,
+	translations,
+}: Props) => {
+	const [searchTerm, setSearchTerm] = useState('');
+	const [sortBy, setSortBy] = useState<SortOption>('alphabetical');
+	const [shuffleSeed, setShuffleSeed] = useState(0);
+	const [selectedRoles, setSelectedRoles] = useState<string[]>([]);
+	const [selectedStatuses, setSelectedStatuses] = useState<PersonStatus[]>(
```
MumenthalerManuel: Now I understand the usage of the status filter earlier. But is this really necessary or could we just pass in "showStatusFilter" as a boolean instead of the "defaultStatus"? If you want to show only inactive users in a grid the filter might not really make sense anyway?

website/src/components/storyblok/shared/person-grid-interactive.tsx
```diff
@@ -0,0 +1,437 @@
...
+					className="h-10 w-10 shrink-0"
+					aria-label={translations.sortAriaLabel}
+				>
+					<ArrowUpDownIcon className="size-4" />
+				</Button>
+			</DropdownMenuTrigger>
+			<DropdownMenuContent align="start">
+				<DropdownMenuItem onSelect={() => setSortBy('alphabetical')}>{translations.sortAlphabetical}</DropdownMenuItem>
+				<DropdownMenuItem onSelect={() => setSortBy('startDate')}>{translations.sortStartDate}</DropdownMenuItem>
+				<DropdownMenuItem onSelect={reshuffle}>{translations.sortRandom}</DropdownMenuItem>
```
MumenthalerManuel: Does this have a realistic usecase or is this just a "fun thing"? I remember having to remove random sort from another component in the past so this might not be "desired" and adds a lot of nor really needed code.

website/src/components/storyblok/shared/person-card-grid.tsx
MumenthalerManuel: Center last row does not only center the last row but each row. If there is only one row the cards is not aliged to the rest of the content.

MumenthalerManuel (changes requested):
The whole person-grid block looks and feels nice but it is a bit too bloated in code. If possible try to remove some of the optional stuff (center last row, filtering for status in the filter dropdowns, random sort...) and try making it a bit simpler for better maintainability and readability. Splitting it into multiple a bit simpler components might also be an option (interactive / non interactive, manual selection of persons, country office).

MumenthalerManuel:
image
Names are cut off on mobile screen sizes

website/src/components/content-blocks/person-grid.tsx
ssandino: Good point. I simplified here a bit: the card size and link options now have a simple booleans (md/sm and link/no link). We can always add the dialog option back later if we need it.

website/src/components/content-blocks/person-grid.tsx
ssandino: new commit with simplified approach

website/src/components/content-blocks/person-grid.tsx
ssandino: I got rid of the max count option for now. If we end up with a concrete use case, we can add it back in a better way—by limiting the displayed grid rather than handling only a few special cases.

website/src/components/storyblok/shared/person-card-grid.tsx
ssandino: For now, I got rid of centerLastRow. Since this is also a feature request for the program card overview, we can implement it once and use the same approach in both places.

website/src/components/storyblok/shared/person-card-grid.tsx
ssandino: I think this was related to the prop names in the carousel component, but that should be resolved now that we’re using booleans.

website/src/components/storyblok/shared/person-card.tsx
ssandino: removing transform-gpu

website/src/components/storyblok/shared/person-card.tsx
ssandino: That's fine. Can be a computer only thing - as it really helps the user to understand the pill

website/src/components/storyblok/shared/person-grid-interactive.tsx
ssandino: only fun fun fun thing. deleted it.

website/src/components/storyblok/shared/person-grid-interactive.tsx
ssandino: not sure if this is sorted, but yes there is the use case to show only inactive volunteers, active volunteers, but also both together (for example from the same country - all active and former volunteers from Germany)

website/src/components/content-blocks/person-grid.tsx
ssandino: new commit solving it (hopefully)

MumenthalerManuel:
image
@ssandino great job with the changes, thank you! 

Two smaller new UI issues came with the last changes.

- The card height now changes if the name is longer / shorter -> maybe make the role appear on a newline in every case so the cards are always the same height? Same issue appears on larger screen
- The fine line of the portrait being visible through the white background has for some reason gotten more noticeable - or is it just on my screen? Might be something we have to fix on all components using that card layout though so feel free to ignore it / not important for this pullrequest.

ssandino:
Role issue: There doesn’t seem to be a perfect solution that preserves the design while also showing the maximum amount of information in the card. In some cases, truncating a role like Management to M… isn’t very useful. For now, I at least made all cards the same height whenever this happens, and I reduced the font size so the role is less often on a new row. Every now and then a role falls onto the next row, but I think users understand why. (it's mostly to blame on the Mumenthaler and Scheidegger with their loooong names). 

Line issue: The thin border bug has been around for a while and seems to appear more prominently on some cards than others. I tried shortly to identify the cause but the changes affected too many unrelated things (and didn't solve it), so I think it’s better to tackle this in a separate issue later as the PR is already big. 

Good for me to ship.

ssandino:
Screenshot 2026-07-28 at 17 11 40

Fine for me

MumenthalerManuel (approved):

PR #2796, opened 2026-08-28
feat: add transparency countries block

nschaer92:
## Summary by CodeRabbit

* **New Features**
  * Added a transparency chart showing contribution totals by country.
  * Added interactive country segments with accessible labels, focus states, and an “other countries” details dialog.
  * Added localized support in English, German, French, and Italian.
  * Improved country flag rendering with decorative and custom styling options.

* **Bug Fixes**
  * Reset flag fallback states correctly when the displayed country changes.

* **Tests**
  * Added coverage for charts, flags, translations, distribution calculations, colors, and transparency data.

### DIFF
--- a/website/src/lib/services/transparency/transparency.service.ts
+++ b/website/src/lib/services/transparency/transparency.service.ts
@@ -1,14 +1,18 @@
 import { type PrismaClient } from '@/generated/prisma/client';
 import { PayoutStatus, type CountryCode } from '@/generated/prisma/enums';
 import { getCountryNameByCode, isValidCountryCode } from '@/lib/types/country';
+import { getCountryFlagColors } from '@/lib/utils/country-flag-colors';
 import { BaseService } from '../core/base.service';
 import type { ServiceResult } from '../core/base.types';
 import { type ReserveReadService } from '../reserves/reserve-read.service';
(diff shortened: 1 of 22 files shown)


website/src/lib/services/transparency/transparency.service.ts
```diff
@@ -170,10 +198,28 @@ export class TransparencyService extends BaseService {
 		);
 	}
 
-	private async getContributionsByCountry(limit: number): Promise<ContributionsByCountry[]> {
+	private async getContributionsByCountry(
+		limit: number,
+		financialPeriod: TransparencyFinancialPeriod,
+	): Promise<ContributionsByCountry[]> {
+		const allCountries = await this.getCountryContributionRows(financialPeriod);
+		const grandTotal = allCountries.reduce((sum, row) => sum + row.totalChf, 0);
```
MumenthalerManuel, line 206: @nschaer92 can you verify that this calculation is correct? the next line fliters the countries aond only takes the (15?) first - what happens if we have 20+ countries with relevant contributions?

MumenthalerManuel (approved):

website/src/lib/services/transparency/transparency.service.ts
nschaer92: image
@MumenthalerManuel The result looks reasonable with a dump of the production data. The 15 countries with the most contributions are displayed. The remaining countries are currently grouped together as the 16th entry.

